PR #1, opened 2016-09-17
Avoid importing modules unless actually used

orent:
This patch helps avoid unnecessarily importing modules such as uuid, fractions, email, etc. The datetime module is still imported for now because of the compat module.

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -1,12 +1,7 @@
 import re
 import struct
 from datetime import datetime, timedelta
-from decimal import Decimal
-from email.parser import Parser
-from fractions import Fraction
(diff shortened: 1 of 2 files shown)


agronholm:
What's the benefit? Do those modules take up a significant amount of memory?

orent:
Memory, loading time, and my personal sense of aesthetics...

agronholm:
I'm curious – how does this affect the loading time?

agronholm:
Oh well, might as well merge (when you get it to pass the checks).

agronholm:
Speaking of which, why did the coverage go down?

agronholm:
Thanks!

PR #2, opened 2016-09-17
Easier way to decode bignum

orent:


cbor2/decoder.py
```diff
@@ -177,10 +177,9 @@ def decode_epoch_datetime(self, value, fp, shareable_index=None):
 
     def decode_positive_bignum(self, value, fp, shareable_index=None):
         # Semantic tag 2
-        if PY2:
-            return sum(ord(b) * (2 ** (exp * 8)) for exp, b in enumerate(reversed(value)))
-        else:
-            return sum(b * (2 ** (exp * 8)) for exp, b in enumerate(reversed(value)))
+        from binascii import hexlify
+        result = int(hexlify(value), 16)
+        return result
```
agronholm: Why do you assign it to a local variable instead of returning it immediately?

orent: It was a silly workaround for the coverage bot complaining.

agronholm:
I have incorporated this change separately and credited you in the version history. Thanks!

PR #5, opened 2016-11-09
Add support for simple value (0xe0 to 0xf3 and 0xf8)

jelu:


agronholm:
FYI, I don't accept any PR that adds code without corresponding tests.
I'm also curious to hear why this is necessary.

jelu:
@agronholm yes yes hold your horses :) hopefully added test correctly

look at https://tools.ietf.org/html/rfc7049#appendix-B and this module seem to have missed simple value 0xe0-0xf3 and 0xf8.

agronholm:
What are they used for? There is not even encoding support for them here. Do other implementations support them?

jelu:
@agronholm yupp, tinycbor (C impl) does and it's a nice way to be able to have different meaning to values, simple value only does up to 255.

jelu:
Let me add encoder support also before you merge.

jelu:
@agronholm sorry for all the rebase&push -f, so what do you think?

agronholm:
Will there be round-trip support? If I have a relay that decodes and encodes CBOR, will the simple values persist?
Also, coverage has dropped from its original 100%. I would like that fixed.

jelu:
Coverage was not 100% to begin with, so that was not my doing. If the relay supports RFC 7049 it should, it's in the spec, simple value is nothing new.

agronholm:
Am I correct in understanding that simple values are meant for passing custom singletons around? The likes of None, True, False etc? If so, a custom type (one like CBORTag) should be used to provide round trip compatibility.

agronholm:
No single Python version can get the coverage to 100% but the combined coverage was at 100%, as you can see here: https://coveralls.io/builds/8740597

jelu:
build 41 was the build before me, coverage 99.21% https://coveralls.io/builds/8589851
my last build 51, coverage 99.23%
it was not 100%!

jelu:
Simple value does not need a tag since it's in the specification, it is not an extension.

agronholm:
I didn't say it needs a tag. I'm asking what it's used for. Is it used for passing around custom singletons? Or what else is it good for?

agronholm:
As for the coverage, you are right – lines 308-309 in `decoder.py` are not covered by tests.

jelu:
You can use it for whatever, I am using it for passing bits maps indicating what values are next. Using a simple value instead of integer I can notice the difference when decoding so for example if I see a simple value I know its a bit map, if I don't I know that either all the values that follows are present or not.

agronholm:
And with these changes, how do you tell a decoded simple value apart from a normal integer?

jelu:
That depend on implementation. For tinycbor you can check what value is coming next and when you decode you get a `CborValue` which tells you what it is. You could add something that checks what the next value is.

agronholm:
I asked how in YOUR implementation in this particular PR can someone tell whether they just got a simple value or an integer.

agronholm:
In other words, how did you design simple values to be used?

jelu:
My stuff is in C using Tinycbor, now I'm writing decoder for my stuff in other languages.

Could add a `CBORSimpleValue` class for this that wraps the value?

agronholm:
Yes, put it in `types`. This is what I was going for :)

jelu:
Ah okey I see, thought you meant an actual CBOR tag before but no. Sorry, first time I'm using this library.

jelu:
@agronholm so, ok now?

agronholm:
Getting there. But please remove the explicit call to `CBOREncoder.encode_simple_value` from the test. It's not supposed to be called directly. And move the validation to the `CBORSimpleType` class.

agronholm:
Just one tiny little flake8 failure in test_encoder.py. Fix that and I'll merge.

agronholm:
Thanks for the contribution!

jelu:
Np, cheers!

PR #7, opened 2017-04-28
infinite -> indefinite

orent:
In some cases the code deals both with infinite values, but this is inDEfinite length (per the RFC)


agronholm:
You're right, thanks!

Issue #6, opened 2016-11-23
Canonical cbor


jkbbwr:
Can this library support canonical encoding

The RFC has some suggestions for canonical encoding, the most important is map ordering. Can this library support the ordering for the maps 

>    o  The keys in every map must be sorted lowest value to highest.
      Sorting is performed on the bytes of the representation of the key
      data items without paying attention to the 3/5 bit splitting for
      major types.  (Note that this rule allows maps that have keys of
      different types, even though that is probably a bad practice that
      could lead to errors in some canonicalization implementations.)
      The sorting rules are:
      *  If two keys have different lengths, the shorter one sorts
         earlier;
      *  If two keys have the same length, the one with the lower value
         in (byte-wise) lexical order sorts earlier.

agronholm:
What do you need this for? I could accept a non-invasive PR.

jkbbwr:
I looked into doing a PR but because you have writing quite tightly bound to encoding it makes sorting the keys very tricky without duplicating code :(

I need this for hashing and ensuring content. 

Take dictionary -> CBOR -> sha256.

agronholm:
Could you write the code in a different manner without compromising speed and flexibility?

jkbbwr:
You could separate the writing and encoding. Means one function call overhead when actually writing bytes? Its a real shame there is no inline directive in Python

Something like 

```
def encode_string(x): 
    return x.encode("utf8")
def write_string(x, fp): 
    fp.write(length("snip")) 
    fp.write(encode_string(x))
```

jkbbwr:
The actual implementation of sorted maps is very simple

```
sorted_keys = sorted(map.keys(), key=lambda x: (len(x), x))
for key in sorted_keys:
    # blah
```

agronholm:
How would the code look like with canonical vs default encoding?

jkbbwr:
The encode method becomes

```
    @shareable_encoder
    def encode_map(self, map, fp):
        fp.write(encode_length(0xa0, len(map)))

        if self.canonical:
            keys = sorted(map.keys(), key=lambda x: (len(x), x))
        else:
            keys = map.keys()

        for key in keys:
            self.encode(key, fp)
            self.encode(map[key], fp)
```

jkbbwr:
This might be a nice solution

```
from functools import singledispatch

import struct


@singledispatch
def encode(x, **kwargs):
    raise TypeError("I don't understand %r" % x)


@encode.register(type(None))
[...]

PR #13, opened 2018-01-23
Added missing canonical encoding. Fix for #6

Sekenre:
Also includes some superficial tests

Not tested:
* Performance
* full round-trip encode decode creating identical data-structures and bytes.

### DIFF
--- a/cbor2/compat.py
+++ b/cbor2/compat.py
@@ -64,53 +64,41 @@ if sys.version_info.major >= 3 and sys.version_info.minor >= 6:
 
 
 else:
-    try:
-        from numpy import float16, dtype, frombuffer
+    from math import ldexp
+    import struct
 
-        big_e_float16 = dtype(float16).newbyteorder('>')
+    def pack_float16(value):
+        # Based on node-cbor by hildjj
+        # which was based in turn on Carsten Borman's cn-cbor
+        u32 = struct.pack('>f', value)
+        u = struct.unpack('>I', u32)[0]
 
-        def pack_float16(value):
-            return b'\xf9' + float16(value).byteswap().tobytes()
+        if u & 0x1FFF != 0:
+            return False
 
-        def unpack_float16(payload):
-            return frombuffer(payload, dtype=big_e_float16)[0]
+        s16 = (u >> 16) & 0x8000
+        exponent = (u >> 23) & 0xff
+        mantissa = u & 0x7fffff
 
-    except ImportError:
-        from math import ldexp
-        import struct
-
-        def pack_float16(value):
-            # Based on node-cbor by hildjj
-            # which was based in turn on Carsten Borman's cn-cbor
-            u32 = struct.pack('>f', value)
-            u = struct.unpack('>I', u32)[0]
-
-            if u & 0x1FFF != 0:
+        if exponent >= 113 and exponent <= 142:
+            s16 += ((exponent - 112) << 10) + (mantissa >> 13)
+        elif exponent >= 103 and exponent < 113:
+            if mantissa & ((1 << (126 - exponent)) - 1):
                 return False
 
-            s16 = (u >> 16) & 0x8000
-            exponent = (u >> 23) & 0xff
-            mantissa = u & 0x7fffff
+            s16 += ((mantissa + 0x800000) >> (126 - exponent))
+        else:
+            return False
+        return struct.pack('>BH', 0xf9, s16)
 
-            if exponent >= 113 and exponent <= 142:
-                s16 += ((exponent - 112) << 10) + (mantissa >> 13)
-            elif exponent >= 103 and exponent < 113:
-                if mantissa & ((1 << (126 - exponent)) - 1):
-                    return False
+    def unpack_float16(payload):
+        # Code adapted from RFC 7049, appendix D
+        def decode_single(single):
+            return struct.unpack("!f", struct.pack("!I", single))[0]
 
-                s16 += ((mantissa + 0x800000) >> (126 - exponent))
-            else:
-                return False
-            return struct.pack('>BH', 0xf9, s16)
+        payload = struct.unpack('>H', payload)[0]
+        value = (payload & 0x7fff) << 13 | (payload & 0x8000) << 16
+        if payload & 0x7c00 != 0x7c00:
+            return ldexp(decode_single(value), 112)
 
-        def unpack_float16(payload):
-            # Code adapted from RFC 7049, appendix D
-            def decode_single(single):
-                return struct.unpack("!f", struct.pack("!I", single))[0]
-
-            payload = struct.unpack('>H', payload)[0]
-            value = (payload & 0x7fff) << 13 | (payload & 0x8000) << 16
-            if payload & 0x7c00 != 0x7c00:
-                return ldexp(decode_single(value), 112)
-
-            return decode_single(value | 0x7f800000)
+        return decode_single(value | 0x7f800000)
--- a/cbor2/compat.py
+++ b/cbor2/compat.py
@@ -47,3 +47,66 @@ else:
     long = int  # noqa
     unicode = str  # noqa
     bytes_from_list = bytes
+
+if sys.version_info.major >= 3 and sys.version_info.minor >= 6:
+    # Python 3.6 added 16 bit floating point to struct
+    import struct
+
+    def pack_float16(value):
+        return struct.pack('>Be', 0xf9, value)
+
+    def unpack_float16(payload):
+        return struct.unpack('>e', payload)[0]
+
+
+else:
+    try:
+        from numpy import float16, dtype, frombuffer
+
+        big_e_float16 = dtype(float16).newbyteorder('>')
+
+        def pack_float16(value):
+            return b'\xf9' + float16(value).byteswap().tobytes()
+
+        def unpack_float16(payload):
+            return frombuffer(payload, dtype=big_e_float16)[0]
+
+    except ImportError:
+        from math import ldexp
+        import struct
+
+        def pack_float16(value):
+            # Based on node-cbor by hildjj
+            # which was based in turn on Carsten Borman's cn-cbor
+            u32 = struct.pack('>f', value)
+            u = struct.unpack('>I', u32)[0]
+
+            if u & 0x1FFF != 0:
+                return False
+
+            s16 = (u >> 16) & 0x8000
+            exponent = (u >> 23) & 0xff
+            mantissa = u & 0x7fffff
+
+            if exponent >= 113 and exponent <= 142:
+                s16 += ((exponent - 112) << 10) + (mantissa >> 13)
+            elif exponent >= 103 and exponent < 113:
+                if mantissa & ((1 << (126 - exponent)) - 1):
+                    return False
+
+                s16 += ((mantissa + 0x800000) >> (126 - exponent))
+            else:
+                return False
+            return struct.pack('>BH', 0xf9, s16)
+
+        def unpack_float16(payload):
+            # Code adapted from RFC 7049, appendix D
+            def decode_single(single):
+                return struct.unpack("!f", struct.pack("!I", single))[0]
+
+            payload = struct.unpack('>H', payload)[0]
+            value = (payload & 0x7fff) << 13 | (payload & 0x8000) << 16
+            if payload & 0x7c00 != 0x7c00:
+                return ldexp(decode_single(value), 112)
+
+            return decode_single(value | 0x7f800000)
--- a/cbor2/compat.py
+++ b/cbor2/compat.py
@@ -1,7 +1,9 @@
+from math import ldexp
+import struct
 import sys
 
 
-if sys.version_info.major < 3:
+if sys.version_info.major < 3:  # pragma: no cover
     from datetime import tzinfo, timedelta
 
     class timezone(tzinfo):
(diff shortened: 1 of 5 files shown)


tests/test_encoder.py
```diff
@@ -258,3 +259,29 @@ def test_dump_to_file(tmpdir):
...
+    'byte length',
+    'integer keys'])
+def test_ordered_map(value, expected):
+    expected = unhexlify(expected)
+    assert dumps(value, canonical=True) == expected
+
+
+def test_map_exception():
+    with pytest.raises(CBOREncodeError) as e_info:
+        dumps({Decimal(33.3): u''}, canonical=True)
```
agronholm: Probably not a good idea to initialize a Decimal with a float (in general).

tests/test_encoder.py
```diff
@@ -258,3 +259,29 @@ def test_dump_to_file(tmpdir):
...
+    s = 'Canonical serialization requires string, bytes or integer keys in a mapping'
+    assert e_info.value.args[0] == s
```
agronholm: I prefer the following:

```python
exc = pytest.raises(CBOREncodeError, dumps, {Decimal('33.3'): u''}, canonical=True)
exc.match('Canonical serialization requires string, bytes or integer keys in a mapping')
```

tests/test_encoder.py
```diff
@@ -258,3 +259,29 @@ def test_dump_to_file(tmpdir):
...
+
+
+@pytest.mark.parametrize('value, expected', [
+    (OrderedDict([(b'a', b''), (b'b', b'')]), 'A2416140416240'),
+    (OrderedDict([(b'b', b''), (b'a', b'')]), 'A2416140416240'),
+    (OrderedDict([(u'a', u''), (u'b', u'')]), 'a2616160616260'),
+    (OrderedDict([(u'b', u''), (u'a', u'')]), 'a2616160616260'),
+    (OrderedDict([(b'00001', u''), (b'002', u'')]), 'A2433030326045303030303160'),
+    (OrderedDict([(255, 0), (2, 0)]), 'a2020018ff00')
+], ids=[
```
agronholm: Put the ids on the same line or wrap on the following lines only when they go over the 99 column limit.

tests/test_encoder.py
```diff
@@ -5,6 +5,7 @@
 from email.mime.text import MIMEText
 from fractions import Fraction
 from uuid import UUID
+from collections import OrderedDict
```
agronholm: Sort imports alphabetically.

Sekenre:
Ok, seems sensible. Thanks for looking at it so quickly.

cbor2/encoder.py
```diff
@@ -114,6 +114,31 @@ def encode_map(encoder, value):
         encoder.encode(val)
 
 
+def encode_sortable_key(val):
+    '''Takes a key and calculates the length of it's optimal byte representation'''
```
agronholm: PEP 8 requires double quotes for docstrings (""" not ''').

cbor2/encoder.py
```diff
@@ -114,6 +114,31 @@ def encode_map(encoder, value):
...
+    if val.__class__ == bytes:
+        return len(val), val
+    if val.__class__ == unicode:
```
agronholm: Add an empty line to separate two distinct `if` blocks.

cbor2/encoder.py
```diff
@@ -114,6 +114,31 @@ def encode_map(encoder, value):
...
+        x = val.encode('utf-8')
+        return len(x), x
+    if val.__class__ == int:
```
agronholm: Same here.

agronholm: I would prefer `if type(val) is int:`. The `__class__` attribute may not even exist on some obscure built-in types. 

This of course applies to the previous lines too.

cbor2/encoder.py
```diff
@@ -114,6 +114,31 @@ def encode_map(encoder, value):
...
+        bytelen = len(hex(abs(val))[1:]) // 2
+        return bytelen, val
+    else:
+        s = 'Canonical serialization requires string, bytes or integer keys in a mapping'
+        raise CBOREncodeError(s)
+
+
+@shareable_encoder
+def encode_canonical_map(encoder, value):
+    '''Reorder keys according to Canonical CBOR specification'''
```
agronholm: Ditto with the docstring.

cbor2/encoder.py
agronholm: "its", not "it's".

cbor2/encoder.py
```diff
@@ -114,6 +114,31 @@ def encode_map(encoder, value):
...
+def encode_sortable_key(val):
+    '''Takes a key and calculates the length of it's optimal byte representation'''
+    if val.__class__ == bytes:
+        return len(val), val
+    if val.__class__ == unicode:
+        x = val.encode('utf-8')
+        return len(x), x
+    if val.__class__ == int:
+        bytelen = len(hex(abs(val))[1:]) // 2
+        return bytelen, val
```
agronholm: This test is insufficient. For example, the number 24 can be represented with a single byte in CBOR while 25 cannot. I'm afraid the code will get a bit more complex.

cbor2/encoder.py
```diff
@@ -240,6 +265,12 @@ def encode_undefined(encoder, value):
     (CBORTag, encode_semantic)
 ])
 
+canonical_encoders = OrderedDict([
```
agronholm: Canonical encoding involves more than just different encoding of dicts. The RFC says that floats must be encoded to the smallest size they can fit into. Your PR does not do this.

cbor2/encoder.py
```diff
@@ -114,6 +114,31 @@ def encode_map(encoder, value):
...
+        return bytelen, val
+    else:
+        s = 'Canonical serialization requires string, bytes or integer keys in a mapping'
+        raise CBOREncodeError(s)
+
+
+@shareable_encoder
+def encode_canonical_map(encoder, value):
+    '''Reorder keys according to Canonical CBOR specification'''
+    keyed_keys = ((encode_sortable_key(key), key) for key in value.keys())
```
agronholm: Couldn't you have simply encoded the keys and then sorted the results? What do you need `encode_sortable_key()` for?

cbor2/encoder.py
```diff
@@ -253,20 +284,26 @@ class CBOREncoder(object):
     :param default: a callable that is called by the encoder with three arguments
         (encoder, value, file object) when no suitable encoder has been found, and should use the
         methods on the encoder to encode any objects it wants to add to the data stream
+    :param canonical: Forces mapping types to be output in a stable order to guarantee that the
+        output will always produce the same hash given the same input. This adds an additional
+        restriction on the datatypes that can be used as keys. In this case only strings, bytes
```
agronholm: Where does this restriction come from? Certainly not the RFC.

Sekenre: Using floats as mapping keys is ambiguous, the spec handwaves it away in the section about Canonical encoding. See: https://www.rfc-editor.org/errata_search.php?rfc=7049

Particularly when length of encoding is more significant for key ordering than magnitude.

Is it a good idea to extend map encoding to every hashable type?

cbor2/encoder.py
Sekenre: Good point, I had assumed we were already using the smallest representation

cbor2/encoder.py
Sekenre: I couldn't just call encoder.encode(key) since that wouldn't return a value, it would just write to the output file...

cbor2/encoder.py
Sekenre: I just copied the style used for dispatching on object types in CBOREncoder https://github.com/agronholm/cbor2/blob/67beccf3c738948d30cc76b7e37e0a350b2185a9/cbor2/encoder.py#L315

cbor2/encoder.py
Sekenre: From the RFC: "Sorting is performed on the bytes of the representation of the key
      data items without paying attention to the 3/5 bit splitting for
      major types."

cbor2/encoder.py
agronholm: Theoretically, when floats are encoded canonically, they should work as keys. Can you accurately do key lookups for floats? Probably not, but it will work for iteration. The RFC does not forbid it so I don't see why we should.

cbor2/encoder.py
agronholm: Sure, but this implementation has no canonicalization whatsoever for floats. If there is another CBOR implementation that does canonicalization, it might be worth taking a look at.

cbor2/encoder.py
agronholm: Right, I keep forgetting that :) The RFC is a bit unclear on how this is supposed to work. Anyhow, I think I can improve on that integer encoding bit.

cbor2/encoder.py
Sekenre: I think this implementation would be the first in the python world. I will add a minimizing float encoder and check other implementations. We should probably warn users that canonical storage of many floats will really kill performance.

For my own protocols I usually decide on a static scaling factor and just encode as an integer to avoid this whole problem.

cbor2/encoder.py
Sekenre: I will think about this more. Look what I can do in python 3.5:

```python
>>> d = { 7.0: "7", 33.3: "33.3" }
>>> d[7]
'7'
>>> d[7.0]
'7'
>>> d[33]
  File "<pyshell#168>", line 1, in <module>
    d[33]
KeyError: 33
>>> d[33.3]
'33.3'
>>> d[33.2999999999999999999999999999999999999999999999999999]
'33.3'
>>> d[33.299999999999999999999999999999]
'33.3'
>>> d[33.299999999999]
  File "<pyshell#172>", line 1, in <module>
    d[33.299999999999]
KeyError: 33.299999999999
>>> 
```

cbor2/encoder.py
Sekenre: i've been playing with this and just calling dumps(value, canonical=True) for every key seems the most straightforward.

agronholm: Yep, that would work for me.

Sekenre:
I'll update the tests once I figure out a 16-bit float encoder

agronholm:
What do you mean by that? This library does have a 16-bit float encoder, can you not adapt that one to your needs?

agronholm:
If I may ask, what is it that you need canonical encoding for?

Sekenre:
I was looking at using cbor as a compact encoding for data archives and message processing where it might pass between several different languages and operating environments. Then I just started having fun tweaking it.

Do you have any suggestions on how I would improve the test coverage on compat.py?

agronholm:
You don't need canonical encoding to pass messages between different languages and operating environments. The only use case I can see for canonical encoding is message hashing.

agronholm:
As for the coverage, I'm not a big fan of the numpy code. You might have to unconditionally declare both code paths for proper testing because the workarounds will get tricky. If you *must* include it, then simulating an `ImportError` might do the trick. If you do include the numpy code in the final version, you'll have to include numpy in the testing dependencies.

Sekenre:
Ok, the numpy stuff I would just use myself for speed, since I usually have that installed anyway. I'll remove it since it complicates things unneccesarily.

cbor2/compat.py
```diff
@@ -47,3 +49,60 @@ def iteritems(self):
     long = int  # noqa
     unicode = str  # noqa
     bytes_from_list = bytes
+
+
+def pack_float16(value):
```
agronholm: Why not have the test on the module level? The test suite is run on 3.6 as well as older Pythons so both code paths should be covered in the end.

cbor2/decoder.py
```diff
@@ -367,7 +357,7 @@ def decode(self, shareable_index=None):
             return decoder(self, subtype, shareable_index)
         except CBORDecodeError:
             raise
-        except Exception as e:
+        except Exception as e:  # pragma: no cover
```
agronholm: This pragma should only be used for inherently untestable code, not circumventing code coverage limits.

agronholm: The question here is, is this case reproduceable? If not, why?

Sekenre: I think it might be testable, I'll keep poking it with a stick.

cbor2/compat.py
Sekenre: will do, I've never used test coverage tools before as you can probably tell!

cbor2/decoder.py
Sekenre: Okeydoke, I can test it like this in python 3:

```python
def test_nested_exception():
    import sys
    if sys.version_info.major >= 3:
        exc = pytest.raises((CBORDecodeError, TypeError), loads, unhexlify('A1D9177082010201'))
        exc.match("error decoding value at index 8: unhashable type: 'CBORTag'")
    else:
        assert True
```

Any objections?

agronholm: Hm, does this really only happen on py3? If so, you should skip the test on py2 (see the pytest documentation on how to do that with a decorator).

Sekenre: Yep I did this in 
https://github.com/agronholm/cbor2/pull/13/commits/c17586f440f816465bb7542d8ff53cf26b5c3d61

agronholm:
Good to see you're still at it :)
Any other changes you want to make before I flip the switch?

cbor2/encoder.py
Sekenre: can this be closed now? I've changed the float encoding to be a bit more efficient...

Sekenre:
Nope, as long as you're happy :)

agronholm:
Thanks!

Issue #15, opened 2018-01-28
CBOR supports Arrays and bytestrings as Map keys, but lists and bytearrays are mutable


Sekenre:
This may be a bit of an edge case, but this breaks any python implementation of `dict = loads('cbor map')` since by default Arrays are interpreted as lists.

An actual use case would be storing Markov Chains which are usually represented as a sequence of tokens as the key, with the following likely tokens stored as a list of values.

Here's a failing test case:

```python
@pytest.mark.parametrize('payload, expected', [
    ('A18201026178', {(1, 2): 'x'})
])
def test_map(payload, expected):
    decoded = loads(unhexlify(payload))
    assert decoded == expected
```

A possible solution is to add an argument to any decoder that usually returns a mutable sequence to let it know that it's being called on a map key and to try and return an immutable type instead.

An even more crazy (and related) edge case is an indefinite length byte-stream being used as a map key, like this: `loads(unhexlify('a15f4100ff01'))`

agronholm:
I would welcome a PR for this.

agronholm:
I take it that the other CBOR implementation has the same issue?

Sekenre:
It appears to, there's an open pull-request for fixing it: https://bitbucket.org/bodhisnarkva/cbor/pull-requests/2/deserialize-arrays-as-tuples-instead-of/diff

agronholm:
Now with set support (tag 285) support in place, sets used as map keys should also be decoded as frozensets.

Sekenre:
Good idea, I may be able to tackle that on Friday...

PR #16, opened 2018-01-29
Fix for #15

Sekenre:
Although encoding a map key as an indefinite length anything seems a bit unlikely I thought I should cover the possibility.


Sekenre:
Closing this to re-submit against latest master.

PR #19, opened 2018-02-23
Raise EOFError if EOF is encountered on read

### LINKED ISSUE #12: Problem with mutliread of cbor2 object ... (generator pattern).
I want to write many objects into one file without use list as proxy since it can be very large amount of data.

How to do it in cbor2 (see cbor works with EOFError pattern)?
It is works in cbor2 looks like bug I get different exception?

```
  File "C:/Users/Cezary Wagner/PycharmProjects/chess-lichess-eval-parse/sandbox/decode_games/s01_pgn_to_cbor.py", line 42, in <module>
    main()
  File "C:/Users/Cezary Wagner/PycharmProjects/chess-lichess-eval-parse/sandbox/decode_games/s01_pgn_to_cbor.py", line 34, in main
    cbor2.load(f)
  File "C:\root\Python36-64\lib\site-packages\cbor2\decoder.py", line 411, in load
  File "C:\root\Python36-64\lib\site-packages\cbor2\decoder.py", line 363, in decode
    .format(self.fp.tell(), e))
cbor2.decoder.CBORDecodeError: error reading major type at index 10126175: index out of range
```

```
import datetime

import cbor2
from cbor import cbor

from cw_chess.perfect_pgn import PerfectPgnGamesOffsetsParser


def main():
    start_time = datetime.datetime.utcnow()
    with open('../parser_eval/data/out-6976.txt', newline='') as pgn, \
            open('data/games_offsets.cbor', 'wb') as f:
        pgn_games_offset_parser = PerfectPgnGamesOffsetsParser(pgn)

        for game_offset in pgn_games_offset_parser.games_offsets():
            cbor.dump(game_offset, f)
    end_time = datetime.datetime.utcnow()
    print(end_time - start_time)

    start_time = datetime.datetime.utcnow()
    with open('data/games_offsets.cbor', 'rb') as f:
        while True:
            try:
                cbor.load(f)
            except EOFError:
                break
    end_time = datetime.datetime.utcnow()
    print(end_time - start_time)

    start_time = datetime.datetime.utcnow()
    with open('data/games_offsets.cbor', 'rb') as f:
        while True:
            try:
                cbor2.load(f)
            except EOFError:
                break
    end_time = datetime.datetime.utcnow()
    print(end_time - start_time)


if __name__ == '__main__':
    main()
```

bazurbat:
This allows to distinguish actual decoding errors from EOF when decoding files containing multiple objects. Also fixes #12.

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -351,11 +351,16 @@ class CBORDecoder(object):
         """
         Decode the next value from the stream.
 
+        :raises EOFError: if EOF is encountered while reading
         :raises CBORDecodeError: if there is any problem decoding the stream
 
         """
+        byte = self.fp.read(1)
+        if not byte:
+            raise EOFError()
+
         try:
-            initial_byte = byte_as_integer(self.fp.read(1))
+            initial_byte = byte_as_integer(byte)
             major_type = initial_byte >> 5
             subtype = initial_byte & 31
         except Exception as e:


Sekenre:
An EOF while decoding should not be used to signal the end of a stream.

If you're decoding a set of CBOR objects that are just concatenated together in a file without delimiters then do this:

```python
from cbor2 import CBORDecoder
with open('myfile.cbor', 'rb') as f:
    decoder = CBORDecoder(f)
    while f.peek(1): # Check that there's data at the next file position
        print(decoder.decode())
    print('Done!')
```

Otherwise wrap the items in an indefinite length array.

Sekenre:
I'm going to close this and open an issue to document how to handle multiple objects in a stream.

PR #21, opened 2018-03-21
Problem: Bug #15, CBOR maps can contain unhashable types as keys

Sekenre:
Solution: Use a context manager for decoding keys and set items which
makes sure to use a hashable type.

Also indefinite length bytestring decoding returns bytes() instead of
bytearray().

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -1,10 +1,11 @@
 import re
 import struct
 from datetime import datetime, timedelta
+from contextlib import contextmanager
 from io import BytesIO
 
-from .compat import timezone, xrange, byte_as_integer, unpack_float16
-from .types import CBORTag, undefined, break_marker, CBORSimpleValue
+from cbor2.compat import timezone, xrange, byte_as_integer, unpack_float16
+from cbor2.types import CBORTag, undefined, break_marker, CBORSimpleValue, HashableMap
 
 timestamp_re = re.compile(r'^(\d{4})-(\d\d)-(\d\d)T(\d\d):(\d\d):(\d\d)'
                           r'(?:\.(\d+))?(?:Z|([+-]\d\d):(\d\d))$')
@@ -43,15 +44,15 @@ def decode_bytestring(decoder, subtype, shareable_index=None):
     length = decode_uint(decoder, subtype, allow_indefinite=True)
     if length is None:
         # Indefinite length
-        buf = bytearray()
+        buf = []
         while True:
             initial_byte = byte_as_integer(decoder.read(1))
             if initial_byte == 255:
-                return buf
+                return b''.join(buf)
             else:
                 length = decode_uint(decoder, initial_byte & 31)
                 value = decoder.read(length)
-                buf.extend(value)
+                buf.append(value)
     else:
         return decoder.read(length)
 
@@ -79,7 +80,10 @@ def decode_array(decoder, subtype, shareable_index=None):
             item = decoder.decode()
             items.append(item)
 
-    return items
+    if decoder._hashable:
+        return tuple(items)
+    else:
+        return items
 
 
 def decode_map(decoder, subtype, shareable_index=None):
@@ -90,7 +94,8 @@ def decode_map(decoder, subtype, shareable_index=None):
     if length is None:
         # Indefinite length
         while True:
-            key = decoder.decode()
+            with decoder.key_decoder() as decode:
+                key = decode()
             if key is break_marker:
                 break
             else:
@@ -98,12 +103,15 @@ def decode_map(decoder, subtype, shareable_index=None):
                 dictionary[key] = value
     else:
         for _ in xrange(length):
-            key = decoder.decode()
+            with decoder.key_decoder() as decode:
+                key = decode()
             value = decoder.decode()
             dictionary[key] = value
 
     if decoder.object_hook:
         return decoder.object_hook(decoder, dictionary)
+    elif decoder._hashable:
+        return HashableMap(dictionary)
     else:
         return dictionary
 
@@ -117,7 +125,13 @@ def decode_semantic(decoder, subtype, shareable_index=None):
         shareable_index = decoder._allocate_shareable()
         return decoder.decode(shareable_index)
 
-    value = decoder.decode()
+    # Special handling for sets
+    if tagnum == 258:
+        with decoder.key_decoder() as decode:
+            value = decode()
+    else:
+        value = decoder.decode()
+
     semantic_decoder = semantic_decoders.get(tagnum)
     if semantic_decoder:
         return semantic_decoder(decoder, value, shareable_index)
@@ -228,7 +242,10 @@ def decode_uuid(decoder, value, shareable_index=None):
 
 def decode_set(decoder, value, shareable_index=None):
     # Semantic tag 258
-    return set(value)
+    if decoder._hashable:
+        return frozenset(value)
+    else:
+        return set(value)
 
 
 #
@@ -304,13 +321,14 @@ class CBORDecoder(object):
         The return value is substituted for the dict in the deserialized output.
     """
 
-    __slots__ = ('fp', 'tag_hook', 'object_hook', '_shareables')
+    __slots__ = ('fp', 'tag_hook', 'object_hook', '_shareables', '_hashable')
 
     def __init__(self, fp, tag_hook=None, object_hook=None):
         self.fp = fp
         self.tag_hook = tag_hook
         self.object_hook = object_hook
         self._shareables = []
+        self._hashable = False
 
     def _allocate_shareable(self):
         self._shareables.append(None)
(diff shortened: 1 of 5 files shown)


cbor2/decoder.py
```diff
@@ -43,15 +44,15 @@ def decode_bytestring(decoder, subtype, shareable_index=None):
     length = decode_uint(decoder, subtype, allow_indefinite=True)
     if length is None:
         # Indefinite length
-        buf = bytearray()
+        buf = []
         while True:
             initial_byte = byte_as_integer(decoder.read(1))
             if initial_byte == 255:
-                return buf
+                return b''.join(buf)
```
agronholm: I'm concerned about returning bytes instead of a bytearray here. Have you checked how much, if any, impact this has on memory use and performance?

cbor2/types.py
```diff
@@ -46,6 +46,15 @@ def __repr__(self):
         return 'CBORSimpleValue({self.value})'.format(self=self)
 
 
+class HashableMap(dict):
```
agronholm: No, just no. This is fundamentally wrong. This is still a mutable type, you're just trying to work around the fact that dicts are not hashable, as they rightfully should not be. Another type could be made where the keys and values cannot be changed but this is not it.

cbor2/decoder.py
```diff
@@ -117,7 +125,13 @@ def decode_semantic(decoder, subtype, shareable_index=None):
         shareable_index = decoder._allocate_shareable()
         return decoder.decode(shareable_index)
 
-    value = decoder.decode()
+    # Special handling for sets
```
agronholm: This code looks out of place. Why is it not in `decode_set()`?

cbor2/decoder.py
```diff
@@ -90,20 +94,24 @@ def decode_map(decoder, subtype, shareable_index=None):
...
         for _ in xrange(length):
-            key = decoder.decode()
+            with decoder.key_decoder() as decode:
+                key = decode()
             value = decoder.decode()
             dictionary[key] = value
 
     if decoder.object_hook:
         return decoder.object_hook(decoder, dictionary)
+    elif decoder._hashable:
```
agronholm: See the comment further down.

cbor2/decoder.py
```diff
@@ -381,6 +399,16 @@ def decode_from_bytes(self, buf):
         self.fp = old_fp
         return retval
 
+    @contextmanager
+    def key_decoder(self):
+        """
```
agronholm: This docstring would fit on one line.

Sekenre:
Totally agree on the HashableMap thing. There are many alternatives implementing frozendict here: https://www.python.org/dev/peps/pep-0416/I am spoiled for choice!On 21 Mar 2018 8:22 p.m., Alex Grönholm <notifications@github.com> wrote:@agronholm requested changes on this pull request.

Some good ideas, some bad.

In cbor2/decoder.py:
>          while True:
             initial_byte = byte_as_integer(decoder.read(1))
             if initial_byte == 255:
-                return buf
+                return b''.join(buf)

I'm concerned about returning bytes instead of a bytearray here. Have you checked how much, if any, impact this has on memory use and performance?



In cbor2/types.py:
> @@ -46,6 +46,15 @@ def __repr__(self):
         return 'CBORSimpleValue({self.value})'.format(self=self)
 
 
+class HashableMap(dict):

No, just no. This is fundamentally wrong. This is still a mutable type, you're just trying to work around the fact that dicts are not hashable, as they rightfully should not be. Another type could be made where the keys and values cannot be changed but this is not it.



In cbor2/decoder.py:
> @@ -117,7 +125,13 @@ def decode_semantic(decoder, subtype, shareable_index=None):
         shareable_index = decoder._allocate_shareable()
         return decoder.decode(shareable_index)
 
-    value = decoder.decode()
+    # Special handling for sets

This code looks out of place. Why is it not in decode_set()?



In cbor2/decoder.py:
>              value = decoder.decode()
             dictionary[key] = value
 
     if decoder.object_hook:
         return decoder.object_hook(decoder, dictionary)
+    elif decoder._hashable:

See the comment further down.



In cbor2/decoder.py:
> @@ -381,6 +399,16 @@ def decode_from_bytes(self, buf):
         self.fp = old_fp
         return retval
 
+    @contextmanager
+    def key_decoder(self):
+        """

This docstring would fit on one line.

cbor2/decoder.py
Sekenre: The values must be hashable before passing to decode_set()

cbor2/decoder.py
Sekenre: I'm sure buffering the string fragments in a list will increase the memory usage. I'm not sure how I could test it's impact.

I believe when using strings, the ''.join(<strings>) method is way more efficient than repeated concatenation. bytearrays probably do it differently though.

Sekenre: I've updated the Gist, at least on python 3.6, calling join on a list of bytes is about 3 times faster and uses half the memory of appending to a bytearray.

Sekenre: If the size of each string is 100 bytes and the number of strings to be concatenated is 500,000 you get:

```python
%%timeit -r 10 -n 100000 data = gen(100,500000)
concat_bytes(data)
```

    100000 loops, best of 10: 14.4 µs per loop
    


```python
%%timeit  -r 10 -n 100000 data = gen(100,500000)
concat_join(data)
```

    100000 loops, best of 10: 12.8 µs per loop
    


```python
%%memit data = gen(100,500000)
concat_bytes(data)
```

    peak memory: 97.41 MiB, increment: 44.68 MiB
    


```python
%%memit data = gen(100,500000)
concat_join(data)
```

    peak memory: 162.55 MiB, increment: 110.62 MiB

Sekenre:
I propose this to replace the abominable HashableMap:

```python
class FrozenDict(Mapping):
    """
    A hashable, immutable mapping type.

    The arguments to ``FrozenDict`` are processed just like those to ``dict``.
    """
    def __init__(self, *args, **kwargs):
        self._d = dict(*args, **kwargs)
        self._hash = None

    @classmethod
    def supply_dict(cls, d):
        """
        When called from decode_map we can safely supply the internal dict
        directly.

        We can guarantee that only immutable pairs have been used to
        construct this dictionary.
        """
        frozen_dict = cls()
        frozen_dict._d = d
        return frozen_dict

    def __iter__(self):
        return iter(self._d)

    def __len__(self):
        return len(self._d)

    def __getitem__(self, key):
        return self._d[key]

    def __repr__(self):
        return "%s(%s)" % (self.__class__.__name__, self._d)

    def __hash__(self):
        if self._hash is None:
            self._hash = hash((frozenset(self), frozenset(self.values())))
        return self._hash
```

cbor2/types.py
Sekenre: Have a look at my proposal below

agronholm:
FrozenDict looks a lot better. I will probably want to tweak it further. For one, I'm not sure if `supply_dict()` is necessary since `dict` can take another dict as an argument.

cbor2/decoder.py
agronholm: I'm not convinced yet but I will take a better look this weekend.

cbor2/decoder.py
agronholm: Why is `_hashable` private but `key_decoder` is public? That makes no sense. If someone wants to implement their own decoder, it would make sense for them to want to read the state of `hashable`, yes?

cbor2/decoder.py
agronholm: Alright, you've convinced me.

cbor2/types.py
```diff
@@ -46,6 +49,35 @@ def __repr__(self):
...
+
+    def __getitem__(self, key):
+        return self._d[key]
+
+    def __repr__(self):
+        return "%s(%s)" % (self.__class__.__name__, self._d)
+
+    def __hash__(self):
+        if self._hash is None:
+            self._hash = hash((frozenset(self), frozenset(self.values())))
```
agronholm: Does the class actually implement the `values()` method?

agronholm: Hm never mind, it seems that the `Mapping` class provides this.

cbor2/decoder.py
```diff
@@ -304,13 +321,14 @@ class CBORDecoder(object):
...
 
-    __slots__ = ('fp', 'tag_hook', 'object_hook', '_shareables')
+    __slots__ = ('fp', 'tag_hook', 'object_hook', '_shareables', '_immutable')
 
     def __init__(self, fp, tag_hook=None, object_hook=None):
         self.fp = fp
         self.tag_hook = tag_hook
         self.object_hook = object_hook
         self._shareables = []
+        self._immutable = False
```
agronholm: My previous comment got lost somewhere, but if decoders must access both `key_decoder` and `_immutable`, why is one public and one private? Shouldn't both be public if third party decoders are allowed?

Sekenre: Hm, I guess it would make sense to make that flag generally global, but you never want to set it to False until the last context manager has exited.

For instance, the object_hook and tag_hook may need to know if it's being passed something contained within a map.

Sekenre: How about this:

```python
@property
def immutable(self):
    return self._immutable
```

Without the corresponding setter method, then the internal flag would only be set by the context manager but it would be readable as:

```python
if decoder.immutable:
    return tuple(value)
```

agronholm: Works for me.

Sekenre:
Let me know if you need more tweaks on this, I might have some time to look at it tomorrow.

agronholm:
I think it's fine, I'll give it another once-over in the near future and merge then.

agronholm:
Thanks!

PR #22, opened 2018-05-27
FrozenDict support was missing from encoder.

Sekenre:
This broke round-trip decode-to-encode if the object used a mapping type as a key in another mapping.


Sekenre:
Could this lead up to tagging a new release? Thanks.

agronholm:
I've released v4.1.0 which includes this addition. Thanks!

Issue #26, opened 2018-10-16
Overflow error encoding large float with canonical=True


bmerry:
```python
>>> cbor2.dumps(1e300, canonical=True)
OverflowError: float too large to pack with f format
```

Using cbor2 4.1.1.

Sekenre:
Fixed in master. Will do a release soon.

agronholm:
@Sekenre Talk to me before you do a release. I want to make sure you know the proper procedure for that.

Sekenre:
@agronholm Sure, You can message me over Linkedin and we could exchange email addresses GPG keys or whatever. Or I'll just '@' you when I'm ready to tag.

PR #27, opened 2018-10-22
Fix for #26: handle overflow exceptions properly...

Sekenre:
...and try a better encoding.

I decided to make it explicit and always try to use the proper struct.pack format and only use the pack/unpack_float16 if there's an exception.

Using canonical encoding with lots of floating point data in python<3.6 will be very slow.

Follows reccommendation in CBOR spec:

https://tools.ietf.org/html/draft-ietf-cbor-7049bis-04#page-32

>   If a protocol allows for IEEE floats, then additional
>   canonicalization rules might need to be added.  One example rule
>   might be to have all floats start as a 64-bit float, then do a test
>   conversion to a 32-bit float; if the result is the same numeric
>   value, use the shorter value and repeat the process with a test
>   conversion to a 16-bit float.  (This rule selects 16-bit float for
>   positive and negative Infinity as well.)  Also, there are many
>   representations for NaN.  If NaN is an allowed value, it must always
>   be represented as 0xf97e00.

### DIFF
--- a/cbor2/compat.py
+++ b/cbor2/compat.py
@@ -55,20 +55,21 @@ if sys.version_info.major >= 3 and sys.version_info.minor >= 6:
     # Python 3.6 added 16 bit floating point to struct
 
     def pack_float16(value):
-        try:
-            return struct.pack('>Be', 0xf9, value)
-        except OverflowError:
-            return False
+        pass
 
     def unpack_float16(payload):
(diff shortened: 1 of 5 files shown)


Sekenre:
Ok Thanks, no hurry, I'm not using canonical in production yet :smiley:

PR #33, opened 2018-11-13
Decoder is wasting a lot of effort on decoding key/value pairs from mappings

Sekenre:
Problem: Enter and exit of with: block takes a lot of time

Solution: If we just set the flag directly it saves 20% of time when
decoding keys from a CBOR mapping.

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -1,7 +1,6 @@
 import re
 import struct
 from datetime import datetime, timedelta
-from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


agronholm:
What was the rationale for the context manager in the first place?

Sekenre:
> 
> 
> What was the rationale for the context manager in the first place?

I was just used to using context managers for managing state in a nice compact way. In retrospect it was a bit ridiculous to use it for a simple boolean flag.

PR #40, opened 2019-02-20
Don't encode dates as datetimes by default

### LINKED ISSUE #37: dates are encoded as datetimes
https://github.com/agronholm/cbor2/blob/bf4acfd4ba08959a8b18f6c97981bae6d9bd22ca/cbor2/encoder.py#L156-L158

In my opinion the default behavior of the encoder should be to refuse to serialize an object if the decoder isn't capable of recreating it exactly. Converting to and from CBOR shouldn't be a lossy operation.

Changaco:
This branch fixes #37 but adds a flag to restore the old behavior. It may not be the most elegant solution, but it's the simplest one (that I know of).

### DIFF
--- a/cbor2/encoder.py
+++ b/cbor2/encoder.py
@@ -66,6 +66,9 @@ class CBOREncoder(object):
         when True, use "canonical" CBOR representation; this typically involves
         sorting maps, sets, etc. into a pre-determined order ensuring that
         serializations are comparable without decoding
+    :param bool date_as_datetime: set to ``True`` to serialize date objects as
+        datetimes (CBOR tag 0), which was the default behavior in previous
+        releases (cbor2 <= 4.1.2).
 
     .. _CBOR: https://cbor.io/
     """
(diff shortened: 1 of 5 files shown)


Changaco:
(CI isn't triggered for draft PRs?)

Sekenre:
I'm happy with this

agronholm:
So long as you remember that this constitutes a breaking API change.

Changaco:
It's my first time writing CPython code, so someone more familiar with CPython should review it thoroughly.

Sekenre:
Thanks for this, I found a difference in behaviour between the C version and Pure python,

If you change your test code as follows you'll see the difference:

```diff
diff --git a/tests/test_encoder.py b/tests/test_encoder.py
index 0eb29d6..db298ec 100644
--- a/tests/test_encoder.py
+++ b/tests/test_encoder.py
@@ -237,7 +237,7 @@ def test_datetime(impl, value, as_timestamp, expected):
 
 def test_date(impl):
     with pytest.raises(impl.CBOREncodeError):
-        impl.dumps(date(2013, 3, 21))
+        impl.dumps(date(2013, 3, 21), timezone=timezone.utc)
 
 
 def test_date_as_datetime(impl):
```

PR #42, opened 2019-03-11
setup: lower setuptools version requirement to 22

pboettch:
This is needed for old packaging system like
Yocto Morty (2016) - in my case.

Signed-off-by: Patrick Boettcher <p@yai.se>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ setup(
         'local_scheme': 'dirty-tag'
     },
     setup_requires=[
-        'setuptools >= 36.2.7',
+        'setuptools >= 22.0.0',
         'setuptools_scm >= 1.7.0'
     ]
 )


agronholm:
Setuptools this old will not properly build the wheel since it can't process the metadata in `setup.cfg`.

agronholm:
The support for `setup.cfg` was added in setuptools v30.3.0.

pboettch:
Ah ok, didn't know. I was just looking for a fix for my problem and wanted to share.

I'll then have to carry a patch with my recipe. One less or more doesn't change anything.

agronholm:
Won't you have to make other changes for the build to work anyway, like moving all the metadata back to `setup.py`?

pboettch:
Maybe, I'll see. 

What was the last version of your lib which would have worked with setuptools before version 30? 

What kind of meta-information from setup.cfg really needed? When reading it, nothing seems to be mandatory to make installation work.

agronholm:
I don't think I ever had a version that worked with such an old setuptools version.
I suggest you look at the history of setup.py and find the commit where I switched to setup.cfg.
If you try to use setuptools 22 to build the egg info, you'll find that the metadata is wrong. The metadata is usually required for Python installers like pip to properly resolve the dependencies when installing other libraries. I don't know what the case is with Yocto though.

pboettch:
AFAICS, Yocto/bitbake is "just" running 'setup.py install' (or similar) and then creates an .ipk-package (similar to .deb) which contains the files. Versions and names are taken from the recipe files and not from the metadata.

Maybe newer version of bitbake with newer setuptools are using the metadata.

PR #45, opened 2019-04-25
Include microseconds with timestamp encoding

### LINKED ISSUE #44: Timestamp encoding throws away microseconds
I get the impression from `encode_datetime` in encoder.py (https://github.com/agronholm/cbor2/blob/master/cbor2/encoder.py#L138) that the intention for timestamp encoding is to calculate a floating-point value to encode. Unfortunately, integer division (// instead of /) is being used with the microseconds field. Given microseconds can never be >1000000 (the value it is being divided by), the result of the division will always be 0, effectively throwing the microseconds data away.

For example, under Python 3.5:

```python3
>>> import cbor2
>>> import datetime as dt
>>> ts = dt.datetime.now()
>>> cbor2.loads(cbor2.dumps(ts, datetime_as_timestamp=1, timezone=dt.timezone.utc))
datetime.datetime(2019, 4, 16, 21, 16, 9, tzinfo=datetime.timezone.utc)
>>> cbor2.loads(cbor2.dumps(ts, datetime_as_timestamp=0, timezone=dt.timezone.utc))
datetime.datetime(2019, 4, 16, 21, 16, 9, 167883, tzinfo=datetime.timezone.utc)
```

waveform80:
Fixes #44 by using float division for microseconds in timestamp
encoding. Also maintains minimal representation (int) when microseconds
field is 0.

### DIFF
--- a/cbor2/encoder.py
+++ b/cbor2/encoder.py
@@ -150,6 +150,7 @@ def encode_datetime(encoder, value):
             timestamp = timegm(value.utctimetuple())
         else:
             timestamp = timegm(value.utctimetuple()) + value.microsecond / 1000000
+
         encode_semantic(encoder, CBORTag(1, timestamp))
     else:
         datestring = as_unicode(value.isoformat().replace('+00:00', 'Z'))
--- a/cbor2/encoder.py
+++ b/cbor2/encoder.py
@@ -1,3 +1,5 @@
+from __future__ import division
+
 import re
 import struct
 from collections import OrderedDict, defaultdict
--- a/cbor2/encoder.py
+++ b/cbor2/encoder.py
@@ -146,7 +146,10 @@ def encode_datetime(encoder, value):
 
     if encoder.datetime_as_timestamp:
         from calendar import timegm
-        timestamp = timegm(value.utctimetuple()) + value.microsecond // 1000000
+        if not value.microsecond:
+            timestamp = timegm(value.utctimetuple())
+        else:
+            timestamp = timegm(value.utctimetuple()) + value.microsecond / 1000000
         encode_semantic(encoder, CBORTag(1, timestamp))
     else:
         datestring = as_unicode(value.isoformat().replace('+00:00', 'Z'))
--- a/tests/test_encoder.py
+++ b/tests/test_encoder.py
@@ -114,9 +114,10 @@ def test_simple_value(value, expected):
      'c07819323031332d30332d32315432323a30343a30302b30323a3030'),
     (datetime(2013, 3, 21, 20, 4, 0), False, 'c074323031332d30332d32315432303a30343a30305a'),
     (datetime(2013, 3, 21, 20, 4, 0, tzinfo=timezone.utc), True, 'c11a514b67b0'),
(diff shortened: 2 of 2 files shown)


agronholm:
LGTM.

agronholm:
I'd be totally comfortable dropping py2 support soon-ish.

Sekenre:
Shall we drop py2 next time we increment the major version number?

waveform80:
> Shall we drop py2 next time we increment the major version number?

I'll weigh in with my (slightly drunken, possibly incoherent) tuppence: it's not my place to suggest what cbor2 should do, but I generally take a pretty strong view on backwards compatibility (most of the libraries I'm involved with like picamera, gpiozero, etc. maintain compatibility with 2.7 and 3.2 onwards ... and most people think I'm crazy for supporting 3.2 but I've still got users on Debian Wheezy who are potentially using it :).

That said, I'd be deeply hypocritical in this case as I've written cboar for compatibility with 3.3+ only. That's largely because cross-major-version compatibility with the CPython API is considerably harder than in pure Python, but also because the main driver for cboar (the piwheels project) is python 3.x only, so I don't really care about 2.x compat there.

Anyway, I'd say in cbor2's case that maintaining 2.x compatibility is *fairly* trivial (it could be simplified a little bit but not much - see below) and probably worth maintaining as a result. That said, I'm also of the opinion that 2.x is ultimately time-limited and that a good rule-of-thumb is to write idiomatic python 3 wherever possible and patch in 2.x compatibility where required, the ultimate goal being that removal of the compatibility hacks should be as simple as possible. For instance, rather than defining compat types for unicode I'd rather re-define str to be unicode (as in 3.x) and use the "bytes" type for byte-strings (included in 3.x and 2.7). There's a nasty, but handy, little hack I tend to use for this which you can see at the start of most gpiozero modules: https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/threads.py#L37 (consider what that line means after `from __future__ import unicode_literals` :).

On how far this can be taken in cbor2: dealing with Python's fundamental types as cbor2 does, this philosophy unfortunately can't be taken anything like as far as I usually do. For instance, in most cases where dict.items() is required (e.g. [here](https://github.com/RPi-Distro/python-gpiozero/blob/f82fefa32de9cfd61b855c78c0c75ff587d2c68c/gpiozero/pins/pi.py#L188)), I'd simply *use dict.items()* because the vast majority of the time I can either:

1. guarantee the dict in question isn't going to be excessively large (who passes a million+ keyword arguments?!) so the difference between items/iteritems in 2.x is negligible or

2. it's a sufficiently fringe case that telling users "move to python 3 if you've got a stupidly huge dict" is reasonable.

However, in cbor2 you've got to handle literally any arbitrary dict on input (i.e. a stupidly huge dict isn't a fringe case) so that's just not an option. Still, just in case it's not obvious from the above, I'd suggest not dropping 2.x compat just yet. Its official EOL is less than a year away and although I'm not even vaguely convinced that 2.x will *actually* disappear quietly after its EOL, it would be a pretty reasonable point for library authors to state to their users "if you want 2.x support, it's on you now".

agronholm:
I'm dropping py2 support in the near future in my other projects, but as maintainer, @Sekenre has the final say here.

PR #48, opened 2019-04-25
Reset shared reference after immutable conversion

### LINKED ISSUE #46: Shared immutable arrays aren't shared after load
Another minor (and very esoteric) issue I came across in the course of writing cboar which is probably best illustrated by example:

```python
>>> import cbor2
>>> a = (1, 2, 3)
>>> b = ((a, a), a)
>>> c = cbor2.loads(cbor2.dumps(b, value_sharing=True))
>>> c
[[[1, 2, 3], [1, 2, 3]], [1, 2, 3]]
>>> c[0][0] is c[0][1]
True
>>> c[0][0] is c[1]
True
>>> c = cbor2.loads(cbor2.dumps(set(b), value_sharing=True))
  File "/home/dave/projects/cbor2/cbor2/decoder.py", line 399, in decode
    return decoder(self, subtype, shareable_index)
  File "/home/dave/projects/cbor2/cbor2/decoder.py", line 142, in decode_semantic
    return semantic_decoder(decoder, value, shareable_index)
  File "/home/dave/projects/cbor2/cbor2/decoder.py", line 253, in decode_set
    return set(value)
TypeError: unhashable type: 'list'

During handling of the above exception, another exception occurred:

  File "/home/dave/projects/cbor2/cbor2/decoder.py", line 431, in loads
  File "/home/dave/projects/cbor2/cbor2/decoder.py", line 403, in decode
    raise CBORDecodeError('error decoding value at index {}: {}'.format(self.fp.tell(), e))
cbor2.decoder.CBORDecodeError: error decoding value at index 21: unhashable type: 'list'
>>> c = cbor2.loads(cbor2.dumps(set(b), value_sharing=False))
>>> c
{((1, 2, 3), (1, 2, 3)), (1, 2, 3)}
```

The reason this occurs is that `decode_array` the function creates the list in which to decode the contents and sets it as the current shareable (assuming a semantic shareable tag has preceded it). However, if immutable is also set (as it will be when encoding the set in the above example) then once the array has been filled it is replaced with a tuple before return, but the *shareable* that was set earlier is *still the list* (not the newly created tuple).

It's probably enough to simply replace the shareable after constructing the tuple (but before returning it). In cboar I've gone a slightly different route but that's because constructing a tuple and a list is rather similar at the C level (and it's a lot more efficient to construct the tuple on the fly than convert a list afterward, assuming the length is known up front).

There's probably a similar issue with `decode_map` when it's outputting a FrozenDict but I haven't come up with a test case for that yet. Anyway, I'll bash together another PR for this later on.

waveform80:
Fixes #46. When decoding immutable shared objects, reset the shared
reference after the conversion of the container to its immutable state
otherwise the container of the immutable item is likely to complain
about hashability.

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -80,9 +80,9 @@ def decode_array(decoder, subtype, shareable_index=None):
             items.append(item)
 
     if decoder.immutable:
-        return tuple(items)
-    else:
-        return items
+        items = tuple(items)
+        decoder.set_shareable(shareable_index, items)
+    return items
 
 
 def decode_map(decoder, subtype, shareable_index=None):
@@ -113,9 +113,10 @@ def decode_map(decoder, subtype, shareable_index=None):
 
     if decoder.object_hook:
         return decoder.object_hook(decoder, dictionary)
-    elif decoder.immutable:
-        return FrozenDict(dictionary)
     else:
+        if decoder.immutable:
+            dictionary = FrozenDict(dictionary)
+            decoder.set_shareable(shareable_index, dictionary)
         return dictionary
 
 
--- a/tests/test_decoder.py
+++ b/tests/test_decoder.py
@@ -251,9 +251,9 @@ def test_uninitialized_shared_reference():
 
 
 def test_immutable_shared_reference():
-    #a = (1, 2, 3)
-    #b = ((a, a), a)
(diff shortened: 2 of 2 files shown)


Sekenre:
Looks good thanks.

PR #51, opened 2019-05-22
Combine cboar and cbor2, close #47

### LINKED ISSUE #47: Combine cboar and cbor2?
Not so much an issue this one as a place for discussion of the possibility, raised in #44, of eventually combining the cboar and cbor2 projects (assuming I get cboar finished and released :).

The case *for* merging:

* All cbor2 users get a decent, free speed boost
* There's very little sacrifice of flexibility (though there is *some* - I've already locked down the initial set of encoders in cboar as an optimization because it led to a substantial speed boost)
* There's no sacrifice of functionality (it's basically the same API, right down to the constructor arguments for CBOREncoder and CBORDecoder, though at the moment I haven't bothered with separate exception classes because I thought that bit of cbor2 was a bit odd - naturally I'd change that if the projects were to merge because backwards compatibility would be a concern)

The case *against* merging:

* A C-based library would be less approachable for people to hack on
* A C-based implementation doesn't have the *fairly* strong guarantee against memory leaks that a pure Python version has
* My original thoughts for cboar were that it would remain an "optimized cbor2", i.e. cbor2 would be its "upstream" defining the functionality and API, and that for 99% of users it would be the library of choice while cboar would exist for the 1% running on weird platforms or with weird requirements who needed the speed and were prepared to sacrifice a little flexibility and the memory leak guarantees

Anyway, I'm happy to be persuaded either way but before anybody at the cbor2 end makes a choice I should probably outline some of the changes I've made in [cboar] (there's not many and they're fairly subtle but they might be important). I'll try and find the time to document more of those tonight (or if you're happy perusing C, go have a look at the code!).

Discuss!

[cboar]: https://github.com/waveform80/cboar

waveform80:
Here we go! Turned out to be a few more than 15 commits... ahem...

As mentioned in the associated ticket (#47) the overall diff is ridiculously huge. I've done a few rebases now to break everything down into (hopefully) manageable / reviewable chunks. A rough overview is as follows:

* The first 4 commits deal with trivial changes to the encoder / decoder.
* The "migrate functions to methods" commit is the first "big" commit which moves most stuff into the CBOREncoder and CBORDecoder classes; although it looks complex it's *mostly* fairly trivial (moving the location of the function and replacing the first arg with "self") but there are *some* slightly more complex changes in there.
* The next "big" commit is the one merging cboar into the package. This was a straight copy of cboar with a couple of regex-based search'n'replaces of CBOAR with CBOR2 etc.
* From here on there's three different types of commits...
* Those labelled "parity work" are commits aimed at bringing the Python and C implementations closer together, i.e. renaming attributes, ensuring exceptions have the same type and message for tests, etc.
* Those labelled "bug" are commits fixing bugs I found while merging the projects - mostly things I'd missed in the cboar tests which the cbor2 tests picked up and on rare occasions the reverse.
* The rest are parts of the merge work itself; making the test suite work across implementations, boosting coverage, including various utility scripts for testing speed, trying to catch ref-leaks, etc.
* Finally, the one thing I haven't looked at in detail yet is the documentation, mostly because I'm not sure how they're currently built (the sphinx Makefile doesn't seem to be there?). That said, they will need adjusting as there's some API changes here (e.g. the change in the `tag_hook` prototype to remove shareable_index, etc).

Most of the commits have a bit more commentary in their commit messages, hopefully enough for some context. If you want anything split up further (or combined, re-ordered, removed, etc.) do let me know. I'm happy rebasing and editing stuff for clarity.

Good luck!


Changaco:
Looks good to me.

Sekenre:
This reads well, thanks for structuring the commits so clearly. I'm going to run some tests and merge hopefully today and tomorrow.

Sekenre:
It doesnt build with Microsofts toolchain which is used by Anaconda on windows. Can you add another filter to setup.py so it doesn't try? I will open another issue for it.

waveform80:
Sure - I must admit I haven't tried this at all under Windows (I don't have a Windows development environment at the moment, but I would like to include the platform if it's not too much trouble). Any obvious error messages?

waveform80:
> It doesnt build with Microsofts toolchain which is used by Anaconda on windows. Can you add another filter to setup.py so it doesn't try? I will open another issue for it.

Don't worry about a new issue - I had a shot at setting up a Windows environment for Python extension building this weekend and I *think* I've got a fix that allows everything to build and work. The issue is mostly the use of the "endian" routines which aren't standard (but they're easy to replace with byte-swapping routines given all the platforms Windows runs on are little-endian).

I should add I was absolutely dreading setting up a Windows environment for this - last time I tried it was a *major* pain, but it looks like things have improved significantly both on the Python and the VS side of things (what with Py3.5 and the "universal" CRT), so this time a straight-forward install of VS2019 Community Edition seemed to include everything I needed (including the Python and Git bits!).

I've just got to tidy up some bits and make sure the test suite still runs properly in both Linux & Windows then I'll push the commits to this branch (should only be a couple more).

waveform80:
Right, there's a few more commits now that get things working on Windows too. Nothing too major, but that last commit took me *ages* to figure out! :)

Sekenre:
everything's working beautifully on Linux, Windows is another story. 🤷‍♂ 

I must have my python dev environment setup all wrong on Windows, it's taken me ages to set it up again from scratch and I'm still getting the same errors.

```
running build_ext
building '_cbor2' extension
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX86\x64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD "-IC:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\include" "-IC:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\include" "-IC:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\winrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\cppwinrt" /Tcsource/module.c /Fobuild\temp.win-amd64-3.6\Release\source/module.obj
module.c
c:\users\sekenre\projects\cbor2\source\encoder.h(17): error C2032: '__timezone': function cannot be member of struct '<anonymous-tag>'
```
Any idea if I'm just doing something wrong? I've tried both Anaconda and the Python install through Visual Studio.

waveform80:
> Any idea if I'm just doing something wrong? I've tried both Anaconda and the Python install through Visual Studio.

My setup was pretty vanilla, but I'm not sure if it's the "right way" of doing things on Windows (I used to code on Windows quite extensively back in my Delphi days, but back then XP was shiny and new, and most of my Windows knowledge has bit-rotted!). Anyway, here's the process I followed:

* Fired up a new VM on Azure, with desktop Windows 10
* Installed VS2019 Community Edition with the "Python Environment" option, and checked the "Python native code" option too (to pull in the full SDK - reported installation size was a little over 10Gb)
* Started VS2019 and selected "Clone or checkout code"
* Selected the "GitHub" option under "Browse a repository"
* Logged into GitHub
* Selected and cloned my cbor2 repo (waveform80/cbor2)
* In the main VS interface now, selected Branches under Team Explorer and switched to the "c_module" branch

Here's where I'm not sure what I'm doing is the "right thing" on Windows, simply because it's what I normally do on Linux (ish):

* Selected Tools / Python / Python Environments to bring up the Python Environments sidebar
* With Python 3.7 (64-bit) selected, clicked "Open in Powershell"
* Switched to the cloned repo (`"cd \Users\dave\source\repos\waveform80\cbor2"`)
* Installed an editable version of library with the "test" option ("pip install -v -e .[test]")
* Ran the test suite ("py.test -v")

I'm guessing that's not the right way of doing things because I couldn't debug interactively within VS to figure out the issue I mentioned above. Still, I figured it didn't matter *that* much because it seemed to build happily, so I assumed Windows compat was done. I'll have a dig into the error you saw a bit later and see if I can reproduce it.

waveform80:
Ah! Just had a quick google - Python [issue 34657] looks relevant. At a guess I just need to rename the "timezone" entry in the `CBOREncoderObject` struct. I'll give that a whirl later.

[issue 34657]: https://bugs.python.org/issue34657

Sekenre:
I followed the same steps for VS on a laptop instead of a vm, and then running `python setup.py build_ext` instead of `pip install -v -e .[test]`

Your googling was much more successful than mine!

waveform80:
Okay, hopefully I've got that patched. I've renamed timezone to tz in the CBOREncoderObject struct (but left the attribute named "timezone" for compatibility - that *should* be fine as the attribute name doesn't get declared in a C struct). I haven't tested this on Windows yet as my setup has Python 3.7 and it looks like the issue ([34657] is a duplicate of [24643]) was fixed in 3.7. I could try and setup a down-level version in another VM, but I figured it might be quicker if you tested it with your already-setup laptop :)

[34657]: https://bugs.python.org/issue34657
[24643]: https://bugs.python.org/issue24643

agronholm:
This sounds like it warrants the use of Azure Pipelines to check for Windows compilation issues.

waveform80:
> This sounds like it warrants the use of Azure Pipelines to check for Windows compilation issues.

Interesting - I'll have a look at this (I was wondering whether it was time to dig into appveyor as well which I think supports Windows too).

agronholm:
Appveyor supports only Windows. Travis currently offers support for Linux, macOS and Windows but their Python support for the latter two is nonexistent. It's particularly awful on macOS. Azure Pipelines has none of these issues and its design makes working across all three operating systems much less of a hassle.

waveform80:
> Appveyor supports only Windows. Travis currently offers support for Linux, macOS and Windows but their Python support for the latter two is nonexistent. It's particularly awful on macOS. Azure Pipelines has none of these issues and its design makes working across all three operating systems much less of a hassle.

Actually Appveyor *claims* to [support Linux](https://www.appveyor.com/docs/linux-images-software/#python) too now: . However, having said that I spent a few hours playing with Azure Pipelines and Appveyor yesterday and couldn't get a thing working on the latter! Even their [own recipe for tox](https://www.appveyor.com/docs/lang/python/) based setups didn't work (on either Linux or Windows) with issues mostly around permission to install stuff (like tox) or upgrade stuff (like setuptools, necessary for recent versions of tox).

By contrast, Azure Pipelines was indeed ridiculously simple: had it up and testing within a few minutes. My one concern with it, is that it demands a fairly scary level of permissions on any repository it's associated with (ability to create commits, branches, etc. etc.) in stark contrast to Appveyor and Travis which both have fairly minimal access requirements. I understand why Pipelines does this - it's trying to provide a very "friendly" interface in which it can set up your repository by creating and committing the necessary configuration - but I must admit I prefer Travis and Appveyor's more minimal approach in this regard.

Anyway, it seems Pipelines is the way to go. There's a couple of branches in my cloned repo that you may want to have a look at:

The PIpelines configuration: https://github.com/waveform80/cbor2/blob/azure-pipelines/azure-pipelines.yml

And my failed Appveyor experiments (in case there's anything blindingly obviously stupid in there): https://github.com/waveform80/cbor2/blob/appveyor/.appveyor.yml

Both of these branch off the `c_module` branch, and here's their respective test portals (hopefully these are public so you should be able to view them):

Azure Pipelines: https://dev.azure.com/waveform88/cbor2

Appveyor: https://ci.appveyor.com/project/DaveJones/cbor2

Happy to take any suggestions for edits to the config; once we've got something reasonably comprehensive across Linux + Windows (+macOS?) I can merge it into the `c_module` branch and then presumably someone can set up PIpelines to work with this repo off that config.

waveform80:
Got Mac OS tests working on Azure Pipelines (I've pretty much given up on Appveyor) and it picked up an issue (surprise surprise) with the endian header again. That's fixed along with a minor issue on 3.5 under Windows (which slightly bizarrely only occurred with one version of VS). Anyway, everything's now passing with CPython 2.7, 3.5, 3.6, 3.7, PyPy, and PyPy3 under Linux, Windows, and Mac OS X (CPython 3.4 is only tested on Linux).

Sekenre:
It works! That was a huge quantity of work, thanks so much!

waveform80:
Wow! Thanks very much for reviewing it!

Do you want me to push the azure pipelines config as another PR? I didn't get around to merging it into `c_module` before it was merged, but it's only a single extra YAML file (azure-pipelines.yml) in the root of the repo.

PR #52, opened 2019-06-05
Azure pipelines

waveform80:
Add a config for Azure Pipelines for CI. This tests the project under Linux, Mac OS X, and Windows (with VS2017 and VS2019), and a variety of interpreter versions.

**NOTE**: Coveralls config is *not* included; this should be added mostly like to the pytest script within the config, and neither is the flake8 config (mostly because I was more concerned with getting the other tests passing :). Deployment isn't included either - I'm not sure how that'd work under Pipelines - it seems to have its own systems for deployment but whether they apply to PyPI I don't know.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,11 +1,16 @@
-# Python package
-# Create and test a Python package on multiple Python versions.
-# Add steps that analyze code, save the dist with the build record, publish to a PyPI-compatible index, and more:
-# https://docs.microsoft.com/azure/devops/pipelines/languages/python
+# Run CI for commits to the master branch, and PRs against the master branch;
+# other branches can be included or excluded here (see Concepts / Triggers in
+# Azure Pipelines docs for more details).
 
 trigger:
 - master
-- c_module
+
+pr:
+- master
+
+
+# Run CI against Linux, Mac OS X, and Windows, and a variety of Python
+# interpreters under each.
 
 strategy:
   matrix:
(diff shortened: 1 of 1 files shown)


agronholm:
You don't really need that many jobs. If @Sekenre allows, we can drop Python 3.4 since it's past its EOL date. For macOS and Windows, we just need the earliest and latest environments unless something warrants having the middle versions tested.

agronholm:
For reference, [here](https://github.com/agronholm/anyio/blob/master/azure-pipelines.yaml) is the latest Azure Pipelines definition I've made.

waveform80:
Ah! Alex's config looks quite a bit more advanced than mine - I'd suggest basing whatever's used on that one and closing this PR. That said, I think it'll still need coveralls adding to it - I had a go with pipelines' built-in coverage handling and wasn't that impressed (it only seemed to keep the last coverage report rather than combining them for overall coverage), but it's entirely possible I missed some config switch somewhere!

Sekenre:
I've never used Python 3.4 nor have I seen it used much. I'm 100% happy to drop it. I made the jump from 2.7 straight to 3.5

waveform80:
3.4's still used in a few places, largely because Debian Jessie is still supported (indeed the download stats on piwheels indicate it's still got a small user-base). But 3.5 (Debian/Raspbian Stretch) is certainly *much* larger and I imagine after Jessie leaves support next year the 3.4 user-base will disappear fairly quickly (Buster with 3.7 will have been out for a while then). Win/Mac platforms tend to be rather more up to date as Python's not (typically) installed from package managers with long-term releases there.

agronholm:
My take is that if somebody is that attached to obsolete software, they can make do with an out of date version of cbor2 too.

Sekenre:
Do I need to have an azure account to see how this works? I'm only really familiar with GitlabCI and a little bit of Travis just from working on this project.

agronholm:
Yes, it requires an Azure account but creating an account is free. I can probably make a PR that adapts the config from my other project.

Issue #56, opened 2019-07-16
Build fails on macOS (various duplicate symbols)


jonashoechst:
When I try to build and install the most recent version of cbor2 on macOS, linking fails because of various duplicate symbol linking errors: 

```
...
clang -bundle -undefined dynamic_lookup -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk build/temp.macosx-10.14-x86_64-3.7/source/module.o build/temp.macosx-10.14-x86_64-3.7/source/encoder.o build/temp.macosx-10.14-x86_64-3.7/source/decoder.o build/temp.macosx-10.14-x86_64-3.7/source/tags.o build/temp.macosx-10.14-x86_64-3.7/source/halffloat.o -L/usr/local/lib -L/usr/local/opt/openssl/lib -L/usr/local/opt/sqlite/lib -lm -o build/lib.macosx-10.14-x86_64-3.7/_cbor2.cpython-37m-darwin.so
duplicate symbol _CBOREncoderType in:
    build/temp.macosx-10.14-x86_64-3.7/source/module.o
    build/temp.macosx-10.14-x86_64-3.7/source/encoder.o
duplicate symbol _CBORTagType in:
    build/temp.macosx-10.14-x86_64-3.7/source/module.o
    build/temp.macosx-10.14-x86_64-3.7/source/encoder.o
duplicate symbol _CBORDecoderType in:
    build/temp.macosx-10.14-x86_64-3.7/source/module.o
    build/temp.macosx-10.14-x86_64-3.7/source/decoder.o
duplicate symbol _CBORTagType in:
    build/temp.macosx-10.14-x86_64-3.7/source/module.o
    build/temp.macosx-10.14-x86_64-3.7/source/decoder.o
duplicate symbol _CBORTagType in:
    build/temp.macosx-10.14-x86_64-3.7/source/module.o
    build/temp.macosx-10.14-x86_64-3.7/source/tags.o
ld: 5 duplicate symbols for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: command 'clang' failed with exit status 1
```

Having a starting point, I think I could help debug the problem.

Sekenre:
Thanks for the PR!

PR #57, opened 2019-08-14
Fix compilation on macOS

jonashoechst:
In order to fix the compilation of cbor2 on macOS, the PyTypeObject variables where marked as `extern` in header files, thus resulting in being linked to the ones defined in the respective c files.


jonashoechst:
... fixes #56.

Issue #60, opened 2019-09-27
requesting support for decoding multiple objects


toravir:
When the logs are in cbor format, there are multiple objects - they are NOT in an array - just objects back to back to back. Example like:

```
{"price":100,"Desc":"Hundred Dollars","level":"info","time":"2019-09-23T11:25:50-07:00"}
{"price":1000,"Desc":"Thousand Dollars","level":"warn","time":"2019-09-23T11:25:50-07:00"}
{"price":1000,"Desc":"Without TS","level":"warn"}
```
load() and loads() just returns the first log line.

One such logger is github.com/rs/zerolog - the binary log format is a CBOR.

waveform80:
I may be missing something here, but while `loads()` will only return the first log line (and it's all I'd *expect* it to return if called with the same input time and again), `load()` most certainly won't provided you don't reset the position of the file-pointer you're passing to it. For example:

```python3
Python 3.5.2 (default, Nov 12 2018, 13:43:14) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cbor2
>>> import io
>>> buf = io.BytesIO()
>>> buf.write(cbor2.dumps({"price":100,"Desc":"Hundred Dollars","level":"info","time":"2019-09-23T11:25:50-07:00"}))
73
>>> buf.write(cbor2.dumps({"price":1000,"Desc":"Thousand Dollars","level":"warn","time":"2019-09-23T11:25:50-07:00"}))
75
>>> buf.write(cbor2.dumps({"price":1000,"Desc":"Without TS","level":"warn"}))
37
>>> buf.seek(0)
0
```

So, at this point we've got a file-like object (buf) containing three log objects (dicts) simply written one after another. We've re-wound the file-pointer to the start, so let's try `load`:

```python3
>>> cbor2.load(buf)
{'Desc': 'Hundred Dollars', 'level': 'info', 'price': 100, 'time': '2019-09-23T11:25:50-07:00'}
>>> cbor2.load(buf)
{'Desc': 'Thousand Dollars', 'level': 'warn', 'price': 1000, 'time': '2019-09-23T11:25:50-07:00'}
>>> cbor2.load(buf)
{'Desc': 'Without TS', 'price': 1000, 'level': 'warn'}
>>> cbor2.load(buf)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #61, opened 2019-09-27
add support for decoding CBOR Sequence

toravir:
Fix for #60.

Added tests..

### DIFF
--- a/cbor2/__init__.py
+++ b/cbor2/__init__.py
@@ -1,4 +1,4 @@
-from .decoder import load, loads, loadMultiple, loadMultiples, CBORDecoder  # noqa
+from .decoder import load, loads, CBORDecoder  # noqa
 from .encoder import dump, dumps, CBOREncoder, shareable_encoder  # noqa
 from .types import (  # noqa
     CBORError,
--- a/cbor2/__init__.py
+++ b/cbor2/__init__.py
@@ -1,4 +1,4 @@
-from .decoder import load, loads, CBORDecoder  # noqa
+from .decoder import load, loads, loadMultiple, loadMultiples, CBORDecoder  # noqa
 from .encoder import dump, dumps, CBOREncoder, shareable_encoder  # noqa
 from .types import (  # noqa
     CBORError,
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -583,17 +583,18 @@ def loadMultiples(s, **kwargs):
     :return:
         array/list of deserialized objects
     """
-    result=[]
+    result =[]
     with BytesIO(s) as fp:
         decObj = CBORDecoder(fp, **kwargs)
         while True:
             try:
                 obj = decObj.decode()
                 result.append(obj)
-            except:
+            except CBORDecodeError:
                 break
 
     return result
+
 
 def loadMultiple(fp, **kwargs):
     """
@@ -606,13 +607,13 @@ def loadMultiple(fp, **kwargs):
     :return:
         array/list of deserialized objects
     """
-    result=[]
+    result = []
     decObj = CBORDecoder(fp, **kwargs)
     while True:
         try:
             obj = decObj.decode()
             result.append(obj)
-        except:
+        except CBORDecodeError:
             break
 
     return result
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -583,7 +583,7 @@ def loadMultiples(s, **kwargs):
     :return:
         array/list of deserialized objects
     """
-    result =[]
+    result = []
     with BytesIO(s) as fp:
         decObj = CBORDecoder(fp, **kwargs)
         while True:
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -570,3 +570,49 @@ def load(fp, **kwargs):
         the deserialized object
     """
     return CBORDecoder(fp, **kwargs).decode()
+
+
+def loadMultiples(s, **kwargs):
+    """
+    Deserialize multiple objects from a bytestring.
+
+    :param bytes s:
+        the bytestring to deserialize
+    :param kwargs:
+        keyword arguments passed to :class:`CBORDecoder`
+    :return:
+        array/list of deserialized objects
+    """
+    result=[]
+    with BytesIO(s) as fp:
+        decObj = CBORDecoder(fp, **kwargs)
+        while True:
+            try:
+                obj = decObj.decode()
+                result.append(obj)
+            except:
+                break
+
+    return result
+
+def loadMultiple(fp, **kwargs):
+    """
+    Deserialize multiple objects from an open file
+
+    :param bytes s:
+        the bytestring to deserialize
+    :param kwargs:
+        keyword arguments passed to :class:`CBORDecoder`
+    :return:
+        array/list of deserialized objects
+    """
+    result=[]
+    decObj = CBORDecoder(fp, **kwargs)
+    while True:
+        try:
+            obj = decObj.decode()
+            result.append(obj)
+        except:
+            break
+
+    return result
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -543,77 +543,48 @@ semantic_decoders = {
 }
 
 
-def loads(s, **kwargs):
+def loads(s, sequence=False, **kwargs):
     """
     Deserialize an object from a bytestring.
 
     :param bytes s:
         the bytestring to deserialize
+    :param bool sequence:
+        set to ``False`` to indicate it is not a CBOR Sequence
+        if set to ``True``, loads will decode multiple CBOR Objects and
+        return a list of decoded objects
     :param kwargs:
         keyword arguments passed to :class:`CBORDecoder`
     :return:
         the deserialized object
     """
     with BytesIO(s) as fp:
-        return CBORDecoder(fp, **kwargs).decode()
+        return load(fp, sequence, **kwargs)
 
 
-def load(fp, **kwargs):
+def load(fp, sequence=False, **kwargs):
     """
     Deserialize an object from an open file.
 
     :param fp:
         the input file (any file-like object)
+    :param bool sequence:
+        set to ``False`` to indicate it is not a CBOR Sequence
+        if set to ``True``, loads will decode multiple CBOR Objects and
+        return a list of decoded objects
     :param kwargs:
         keyword arguments passed to :class:`CBORDecoder`
     :return:
         the deserialized object
     """
-    return CBORDecoder(fp, **kwargs).decode()
-
-
-def loadMultiples(s, **kwargs):
-    """
-    Deserialize multiple objects from a bytestring.
-
-    :param bytes s:
-        the bytestring to deserialize
-    :param kwargs:
-        keyword arguments passed to :class:`CBORDecoder`
-    :return:
-        array/list of deserialized objects
-    """
-    result = []
-    with BytesIO(s) as fp:
-        decObj = CBORDecoder(fp, **kwargs)
-        while True:
-            try:
-                obj = decObj.decode()
-                result.append(obj)
-            except CBORDecodeError:
-                break
-
-    return result
-
-
-def loadMultiple(fp, **kwargs):
-    """
-    Deserialize multiple objects from an open file
-
-    :param bytes s:
-        the bytestring to deserialize
-    :param kwargs:
-        keyword arguments passed to :class:`CBORDecoder`
-    :return:
-        array/list of deserialized objects
-    """
-    result = []
     decObj = CBORDecoder(fp, **kwargs)
-    while True:
-        try:
-            obj = decObj.decode()
-            result.append(obj)
-        except CBORDecodeError:
-            break
+    if not sequence:
+        return decObj.decode()
+
+    result = []
+    while fp.read(1) != '':
+        fp.seek(-1, 1)
+        obj = decObj.decode()
+        result.append(obj)
 
     return result
--- a/source/module.c
+++ b/source/module.c
@@ -316,11 +316,31 @@ CBOR2_load(PyObject *module, PyObject *args, PyObject *kwargs)
 {
     PyObject *ret = NULL;
     CBORDecoderObject *self;
+    bool isSeq = false;
+    PyObject *new_args = args;
 
+    if (kwargs) {
+        PyObject *s = PyDict_GetItem(kwargs, _CBOR2_str_sequence);
+        if (s) {
+            if (PyObject_IsTrue(s))
+                isSeq = true;
+            PyDict_DelItem(kwargs, _CBOR2_str_sequence);
+        }
+    }
     self = (CBORDecoderObject *)CBORDecoder_new(&CBORDecoderType, NULL, NULL);
     if (self) {
-        if (CBORDecoder_init(self, args, kwargs) == 0) {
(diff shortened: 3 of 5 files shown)


Changaco:
I'm not convinced by the proposed API.

A stream of CBOR objects is officially called a CBOR Sequence ([draft-ietf-cbor-sequence-02](https://tools.ietf.org/html/draft-ietf-cbor-sequence-02)), so we should use the word "sequence" instead of "multiple".

The `json` module doesn't have builtin support for JSON sequences, so we can't copy its API.

toravir:
> I'm not convinced by the proposed API.
> 
> A stream of CBOR objects is officially called a CBOR Sequence ([draft-ietf-cbor-sequence-02](https://tools.ietf.org/html/draft-ietf-cbor-sequence-02)), so we should use the word "sequence" instead of "multiple".
> 
> The `json` module doesn't have builtin support for JSON sequences, so we can't copy its API.

would renaming the apis as loadSequence() / loadSequenceStr() be clearer ??

agronholm:
In any case the function names should be named using underscores instead of pascalCase.

Changaco:
Maybe `load_sequence` and `loads_sequence`? Or `load_seq` and `loads_seq`?

Alternatively we could add a `sequence=False` keyword argument to the existing functions. The impact on performance of a single boolean check should be negligable.

toravir:
> Maybe `load_sequence` and `loads_sequence`? Or `load_seq` and `loads_seq`?
> 
> Alternatively we could add a `sequence=False` keyword argument to the existing functions. The impact on performance of a single boolean check should be negligable.

yeah i modifying existing is good with sequence=False.. default. will make changes for it and improve coverage and push those changes tomm..

toravir:
@Changaco  @agronholm - i am new to CPython - so am having trouble writing the logic for sequence in CPython - i have pushed a set of changes from whatever i know - it does not work.. can i get some pointers/ideas please ?

Changaco:
I'm not a CPython expert either, but I managed to get it work. I've opened a PR on @toravir's fork to merge my changes into this branch: <https://github.com/toravir/cbor2/pull/1>.

toravir:
> I'm not a CPython expert either, but I managed to get it work. I've opened a PR on @toravir's fork to merge my changes into this branch: [toravir#1](https://github.com/toravir/cbor2/pull/1).

Thanks @Changaco !! Will look at why the coverage decreased.. at first glance seems unrelated to our changes.

agronholm:
I believe that my method of coverage data gathering might be somewhat incorrect. Ideally, coverage should gather reports from all the test runs and then send the aggregated data to Coveralls. Travis makes this fairly difficult since it does not have good support for passing artifacts from one build stage to the next.

toravir:
> I believe that my method of coverage data gathering might be somewhat incorrect. Ideally, coverage should gather reports from all the test runs and then send the aggregated data to Coveralls. Travis makes this fairly difficult since it does not have good support for passing artifacts from one build stage to the next.

You are right @agronholm , i can't find a way to improve coverage to pass this - can you still merge this PR ?

Sekenre:
Have supplied an alternative implementation in https://github.com/toravir/cbor2/pull/2

Please have a look.

Changaco:
@Sekenre I don't see the point of decoding a sequence as an indefinite length array, it only makes the code more complicated.

@toravir Did you consider yielding the decoded objects instead of adding them all to a list and returning that?

Sekenre:
> @Sekenre I don't see the point of decoding a sequence as an indefinite length array, it only makes the code more complicated.

Meh, it looks simpler to me :smiley: 
> 
> @toravir Did you consider yielding the decoded objects instead of adding them all to a list and returning that?

I would like to add iterator protocol support to the decoder for this exact reason, but having loads return an iterator would break the API.

Also, nesting generators wouldn't work.

maybe adding an `__iter__` method for the `CBORDecoder` object would look the nicest.

Changaco:
@Sekenre A CBOR sequence isn't an indefinite length array, it's not simpler to pretend that it is.

> I would like to add iterator protocol support to the decoder for this exact reason, but having loads return an iterator would break the API.

Not if it only yields when `sequence` is `True`.

toravir:
See #60 for comments - this PR is not needed. The caller can provide an appropriate input in the form of a stream of bytes and the load() will decode each object one by one. With an iterator and an end of file check on the client - multiple CBOR Objects can be decoded withOUT this PR. I will close this PR in 2 days. Just in case someone has something to add/comment.

Changaco:
The fact that the `load` function can be used to read a sequence is pretty much a bug, it should throw an exception like its JSON equivalent:

```python
>>> from io import BytesIO
>>> import json
>>> f = BytesIO(b'{}{}')
>>> json.load(f)
  File "/usr/lib/python3.7/json/__init__.py", line 296, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/usr/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.7/json/decoder.py", line 340, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 1 column 3 (char 2)
```

If you close this PR I will probably open another one.

Sekenre:
> The fact that the `load` function can be used to read a sequence is pretty much a bug, it should throw an exception like its JSON equivalent:
> 

This is actually mentioned in the CBOR Spec: https://tools.ietf.org/html/rfc7049#section-3.2

The spec also mentions streaming decoders. My favourite option is to allow the top-level `CBORDecoder` object to be called as a context-manager and that it returns a generator that yields, similar to a database cursor.

waveform80:
> The fact that the load function can be used to read a sequence is pretty much a bug, it should throw an exception like its JSON equivalent:

Sorry, but this is plain wrong. First, let's check out a couple of other venerable codecs in Python:

```python3
import cbor2
import pickle
import marshal
import io

codecs = (
    pickle,
    marshal,
    cbor2,
)

examples = [
    {"price":100,"Desc":"Hundred Dollars","level":"info","time":"2019-09-23T11:25:50-07:00"},
    {"price":1000,"Desc":"Thousand Dollars","level":"warn","time":"2019-09-23T11:25:50-07:00"},
    {"price":1000,"Desc":"Without TS","level":"warn"},
]

for codec in codecs:
    print(codec.__name__, ':')
    stream = io.BytesIO()
    for example in examples:
        codec.dump(example, stream)
    stream.seek(0)
    try:
        for example in examples:
            decoded = codec.load(stream)
            print(repr(decoded))
            assert decoded == example
    except Exception as e:
        print(str(e))
```

So, marshal, pickle, and cbor2 all produce the same effect. We can simply stream stuff out, and stream it back in again. Why is this possible when JSON throws an exception? Because CBOR is a *framed* format, as are marshal and pickle, unlike JSON which is not. In other words, it's a format in which we can definitely tell when we've come to the end of an element without needing to read any bytes from the next element.

You may look at typical JSON output, in particular your examples above, and say "but we can easily tell when an element has ended: just look for the closing brace". That's true in this *specific* example, but it's not the general case. Check the definition on https://json.org/ and you'll see that "json" is composed of "element", which is composed of "whitespace value whitespace", and "value" can be ... well, pretty much anything. This poses a problem. Consider the following two consecutive writes to JSON output:

```python3
>>> import json
>>> import io
>>> buf = io.StringIO()
>>> buf.write(json.dumps(1))
1
>>> buf.write(json.dumps(1))
1
>>> buf.seek(0)
0
>>> buf.getvalue()
'11'
>>> json.load(buf)
11
```

Ooops. Because there's no framing, two valid representations concatenated together have become a single, different, valid representation. Let's see the same in CBOR:

```python3
>>> import cbor2
>>> import io
>>> buf = io.BytesIO()
>>> buf.write(cbor2.dumps(1))
1
>>> buf.write(cbor2.dumps(1))
1
>>> buf.seek(0)
0
>>> buf.getvalue()
b'\x01\x01'
>>> cbor2.load(buf)
1
>>> cbor2.load(buf)
1
```

How does it know to stop after the first element? Because the length of the element is (in this case implicit) in the encoding. Thus, it doesn't read further than it needs to from the stream.

I realize that streaming JSON is relatively popular these days, but JSON is not a framed format and is *not* suitable for streaming in the general case. In the specific cases that people tend to use (concatenated objects) it's basically a hack, and not a very good one IMHO. Most JSON codecs will (quite reasonably) read everything they can from a stream (because the format explicitly declares it can have arbitrary trailing whitespace) meaning most JSON codecs simply won't work "out of the box" with such streams (not without adding some form of encapsulation, or hacking in some form of "rewind" to push erroneously read characters back into the stream).

Incidentally, another interesting format to look at (in terms of framing), is YAML. This has *optional* framing, which in some ways is even worse. It solves explicitly delimiting frames, but a codec can't assume the framing is in place so it still needs to read arbitrary trailing characters which means it still breaks in the case of throwing the codec on top of a socket (or pipe, etc).

Anyway, I'm afraid in this case, CBOR2 is doing *exactly* what a codec for a framed format is supposed to do: load a single element from the stream leaving the position just beyond the end of that element, permitting future calls to read the next element. Likewise, JSON is doing the only thing a codec for a non-framed format can do: read everything and blow up when it comes across unexpected trailing characters (leaving the caller to blow or up, or try and push the characters back into the stream ... if that's what they want).

Sekenre:
Thanks @waveform80 for the detailed breakdown. If we can keep the surprises to a minimum when people try different serialization methods that's my preferred direction.

From your answer it's clear that JSON is the odd man out and we should not try to exactly emulate it's decoding behaviour.

Changaco:
@waveform80 Your claim about JSON decoding is wrong. The `json.load()` function could stop at the end of the first element it finds, but it's designed to read a file containing a single element, so it raises an exception instead if it finds extra data.

I hadn't thought of looking at pickle, it's true that its load function reads a single object and ignores the rest of the data. In my opinion that's not a good API, but at least this behavior is documented, whereas cbor2's documentation doesn't say anything.

waveform80:
>> If a JSON codec were designed in the manner you're hinting at, it would obviously return the first object complete, but what should it return upon being told to resume reading the stream?

> It would raise an exception complaining that the input is ambiguous.

But it can't do that because the input isn't ambiguous: "111" is perfectly valid JSON (it's just not what we intended), and it has no way of knowing it should be expecting a stream of objects.

You might argue that because the first call returned an object, the later call should expect another object, but the later call to "load" has no knowledge of the prior call to "load". All it knows is it's being passed a file-like object, and that the first byte that's read is "1". That implies the start of a number, possibly the end but we don't know so it'll keep reading the next "1", and so on.

Having load "remember" prior calls is pretty obviously a terrible idea, so let's try another tack. Perhaps "load" should determine whether it's at the start of the file-like object it's been passed? However, in the simplest case that prevents us from using the decoder to read JSON from a socket or a pipe (as noted in the prior post, the "json.load" docs do in fact only mention files ... but personally, I consider that a documentation error).

What about a more complex behaviour? If it can't "tell()" where it is (as in the socket scenario) assume the file-like object is a single document (this is the current behaviour). However, if it *can* tell its position, assume we're in a streaming scenario and only permit top-level objects to be read. Unfortunately, this precludes using the decoder to read encapsulated JSON. And that's besides the obvious fact that it's now operating differently with sockets and files, breaking the "file-like object" abstraction.

> Ideally all the load implementations would behave in the same way regardless of the format being decoded, so as to avoid surprises and subtle bugs.

In an ideal world that would be desirable, but marshal and pickle predate JSON by a long way and already established how "load()" should operate. CBOR merely continues that tradition. JSON is the aberration in this case; the "json" module in Python simply does the best it can with an unframed protocol.

Changaco:
> But it can't do that because the input isn't ambiguous: "111" is perfectly valid JSON (it's just not what we intended), and it has no way of knowing it should be expecting a stream of objects.

The JSON decoder could have been designed to raise an exception on the first call to `load`, complaining that it expected either a whitespace or the EOF, but got a `1` instead. Alternatively it could have been designed to accept any valid input and leave it to the library user to avoid streaming integers without separating them by whitespace. Either way, your claim that the current behavior of the JSON is the only possible one is wrong.

waveform80:
We may have to agree to disagree on this, but I'll try once more:

>> But it can't do that because the input isn't ambiguous: "111" is perfectly valid JSON (it's just not what we intended), and it has no way of knowing it should be expecting a stream of objects.

> The JSON decoder could have been designed to raise an exception on the first call to load, complaining that it expected either a whitespace or the EOF, but got a 1 instead.

"1" is perfectly valid input on the first call to load for any parser compliant with the JSON spec. Absent any extra configuration limiting the parser to a subset of JSON, it cannot reject "1" (or "11" etc.) at the top-level. Or to put it another way, if a parser purported to handle JSON, yet raised an exception upon encountering a "1" on its first read without any further configuration, I'd suggest it doesn't support the JSON spec as written.

> Alternatively it could have been designed to accept any valid input and leave it to the library user to avoid streaming integers without separating them by whitespace.

It could be designed in such a fashion, but there is an issue with this: due to the design of the JSON spec. a parser needs to read-ahead (by at least one char) to know whether it has parsed a complete JSON document. Returning to the example:

    {"foo": 1}111

Our hypothetical parser reads "{" from the input. From the spec. at https://json.org, at this point we know the first "ws" component of "element" is the blank-string, and that we are going down the "object" branch of the "value" token. We proceed with reading "foo": 1 to parse "member", then read "}" which completes "members", thus "object", thus "value". We're now back at "element"; we've accepted the first "ws" and now "value", but according to the grammar there's another "ws" to parse. We read another char from the input and get "1".

Now, at this point you're absolutely correct that the parser could accept blank-string for the final "ws" and return the completed JSON object, leaving a future read to deal with the following digits. But before we do that we've got to rewind that final read. If the input supports seeking (i.e. it's a file), that's fine ... but if the input is non-seekable (a socket, a pipe, etc.) then we've suddenly got a problem.

Either we've got to track state between calls (which then breaks all sorts of things like multiplexing, embedding, etc.), or we could declare that the only valid input is a seekable one (which pretty much limits it to files and nothing else). To be entirely fair ... this is more or less what the Python docs state for [json.load](https://docs.python.org/3/library/json.html#json.load):

> ... Deserialize fp (a .read()-supporting text file or binary file containing a JSON document) to a Python object using this conversion table.

But I can't say I'm entirely convinced that it isn't meant to read "file-like object" rather than just "file". After all, it does only mentioned the necessity that the object supports "read()"; it says nothing of requiring "seek()".

> Either way, your claim that the current behavior of the JSON is the only possible one is wrong.

I didn't exactly claim it's the only possible one; for instance I did include:

> What if the JSON codec was explicitly told to expect only objects, e.g. by some hypothetical argument like json.load(fp, obj_stream=True)? In this case it might well be reasonable to stream JSON. But when operating in such a mode, it could hardly be said to conform to the JSON spec, as it'd reject everything except an object at the top level (granted, that probably wouldn't be of much concern to any users).

However, I would suggest (given the exploration of read-ahead above) that it's the only valid behaviour for a codec that:

1. accepts any file-like object (including non-seekable) as input (like all the other codecs in the stdlib)
2. has been given no further configuration to tell it to expect streaming or some sub-set of the JSON spec.

Now, if either of those limitations are invalid, then anything's possible. However, in assuming both of these, I have to stand by my assertion.

Sekenre:
Closing this since I'm working on an implementation that works a bit differently. Great discussion though!

PR #64, opened 2019-10-29
My own solution to the streaming Iterator problem in #61

Sekenre:
Trying to make the interface as "Pythonic" as possible.

### DIFF
--- a/cbor2/__init__.py
+++ b/cbor2/__init__.py
@@ -1,4 +1,4 @@
-from .decoder import load, loads, CBORDecoder  # noqa
+from .decoder import load, loads, CBORDecoder, CBORStreamDecoder  # noqa
 from .encoder import dump, dumps, CBOREncoder, shareable_encoder  # noqa
 from .types import (  # noqa
     CBORError,
@@ -43,5 +43,6 @@ else:
             ), getattr(_cbor2.CBOREncoder, method.__name__))
             for type_, method in canonical_encoders.items()
         ])
+        CBORStreamDecoder._decoder_class = _cbor2.CBORDecoder
     _init_cbor2()
     del _init_cbor2
--- a/cbor2/__init__.py
+++ b/cbor2/__init__.py
@@ -1,4 +1,4 @@
-from .decoder import load, loads, CBORDecoder, CBORStreamDecoder  # noqa
+from .decoder import load, loads, CBORDecoder, StreamDecoder  # noqa
 from .encoder import dump, dumps, CBOREncoder, shareable_encoder  # noqa
 from .types import (  # noqa
     CBORError,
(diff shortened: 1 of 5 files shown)


Changaco:
Adding an unnecessary context manager doesn't make the code more "Pythonic".

Let's compare the various proposals.

The current way:

```python
import cbor2

with open('myfile.cbor', 'rb') as f:
    while f.peek(1):
        item = cbor2.load(f)
        print(item)
```

Your solution:

```python
from cbor2 import CBORStreamDecoder

with open('myfile.cbor', 'rb') as f:
    with CBORStreamDecoder(f) as cursor:
        for item in cursor:
            print(item)
```

The solution provided by #61:

```python
import cbor2

with open('myfile.cbor', 'rb') as f:
    for item in cbor2.load(f, sequence=True):
        print(item)
```

Which one would you say is more "Pythonic"?

waveform80:
> Adding an unnecessary context manager doesn't make the code more "Pythonic".

I'm rather in agreement with @Changaco here; given the existing implementation happily handles streaming anyway, and in the same manner as existing codecs in Python, I'd just leave it as is. I don't have a strong objection to adding something for those desperate to make streaming explicit ... but on the two occasions I've streamed CBOR it's been over a socket and the procedure has been as simple as:

#### rx.py
```python
import socket
import cbor2

s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
s.bind('/tmp/cbor-stream')
s.listen()
t = s.accept()[0]
f = t.makefile('rb')
while True:
    try:
        print('Received %r' % cbor2.load(f))
    except:
        # Will only happen when the remote end closes
        break
```

#### tx.py
```python
import socket
import cbor2

s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
s.connect('/tmp/cbor-stream')
f = s.makefile('wb')
i = 0
while True:
    input('Press enter to send %d' % i)
    cbor2.dump(i, f)
    f.flush()  # force the send now
    i += 1
```

The one place where cbor2 differs from the built-in codecs (marshal, pickle) is that discussed in #12; i.e. that a different exception is raised at EOF (cbor2 raises CBORDecoderError rather than EOFError). I may have a compromise that would fix that without (seriously) affecting the API, but I'll stick that under #12 where it belongs...

Sekenre:
Hi @waveform80,

This may be a silly question: If you get a partial read from your socket in rx.py and load fails with a decode exception, where does your file pointer end up?

waveform80:
You don't get a partial read with blocking sockets; if you request to read 4 bytes and four bytes aren't on the wire it simply blocks until there are. The only way you get a partial read (i.e. that a read reports EOF) is if the remote end closes the connection (or the connection otherwise fails and is declared closed).

Anyway, good ol' fashioned blocking sockets may not be so popular at the moment (everyone seems to want async and non-blocking for everything), but they do make simple things simple :)

Sekenre:
Hi @Changaco, my thought process wen't like this:

- When streaming i don't want to load the whole file into a list
- load, and loads, should (my preference) always return a complete object or raise an error. A potentially infinite generator does not allow that.
- I didn't want to modify the C decoder (wrapping it is far simpler)
- With this solution it's easier to add features like passing in a filename or a socket instance

Sekenre:
I learned a lot from this but I'm closing it as it's not ideal and/or potentially confusing. I'm working on documenting the alternatives thoroughly.

PR #65, opened 2019-11-01
Close #12

### LINKED ISSUE #12: Problem with mutliread of cbor2 object ... (generator pattern).

waveform80:
Use multiple inheritance to ensure there's still a library-specific root exception, but that exceptions raised by the codec also descend from "typical" Python exceptions like `ValueError` and `TypeError`.

Technically, this is a backwards incompatible change, because I've altered the root exception (`CBORError`) to descend from `Exception` rather than `ValueError`. However, excepting two cases, all exceptions raised encoding and decoding continue to descend from `ValueError`, but now via `CBOREncodeValueError` or `CBORDecodeValueError` rather than `CBORError`.

The two other cases are `CBORDecodeEOF` which now descends from `EOFError` (previously it descended from `ValueError` via `CBORError`), and `CBOREncodeTypeError` for non-serializable types. These are backwards incompatible changes but hopefully they're sufficiently minimal (and sane) that they don't cause any problems.

Incidentally if people think this is going too far, I'm happy to revise the PR to revert the base-class change - just thought I'd present an "idealized" situation first and see if it's sufficiently compatible to keep people happy!

### DIFF
--- a/cbor2/__init__.py
+++ b/cbor2/__init__.py
@@ -3,7 +3,11 @@ from .encoder import dump, dumps, CBOREncoder, shareable_encoder  # noqa
 from .types import (  # noqa
     CBORError,
     CBOREncodeError,
+    CBOREncodeTypeError,
+    CBOREncodeValueError,
     CBORDecodeError,
+    CBORDecodeValueError,
+    CBORDecodeEOF,
     CBORTag,
     CBORSimpleValue,
     undefined
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -6,8 +6,8 @@ from io import BytesIO
 
 from .compat import timezone, range, byte_as_integer, unpack_float16
 from .types import (
-    CBORDecodeError, CBORTag, undefined, break_marker, CBORSimpleValue,
-    FrozenDict)
+    CBORDecodeValueError, CBORDecodeEOF, CBORTag, undefined, break_marker,
+    CBORSimpleValue, FrozenDict)
 
 timestamp_re = re.compile(r'^(\d{4})-(\d\d)-(\d\d)T(\d\d):(\d\d):(\d\d)'
                           r'(?:\.(\d+))?(?:Z|([+-]\d\d):(\d\d))$')
@@ -130,7 +130,7 @@ class CBORDecoder(object):
         """
         data = self._fp_read(amount)
         if len(data) < amount:
-            raise CBORDecodeError(
+            raise CBORDecodeEOF(
                 'premature end of stream (expected to read {} bytes, got {} '
                 'instead)'.format(amount, len(data)))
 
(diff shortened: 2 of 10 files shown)


waveform80:
P.S. I haven't updated the docs in this PR because I've got another branch in which I've nearly finished a tidy-up of the docs. I hope to have that one sorted out over the weekend now but it's turned into a bigger changeset than I'd anticipated; if anyone wants a sneak peak it's under the docs-fixes branch under https://github.com/waveform80/cbor2

Sekenre:
Had a look over this and your in-progress docs. Looks good!

Is there also a test to make sure we're raising each exception in each case?

waveform80:
> Is there also a test to make sure we're raising each exception in each case?

Hmm, good point - the existing tests do cover *some* cases but I don't think there's anything specific for the new derivatives or things like EOF. I'll bung another commit on here with some specific tests.

Issue #67, opened 2020-01-16
wheel fails to build for 4.2.0 on python 3.6.1


patr1ckm:
```
  Running setup.py bdist_wheel for cbor2 ... - \ error
  Complete output from command /home/circleci/project/venv/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-af288b_v/cbor2/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/tmped_mpc5xpip-wheel- --python-tag cp36:
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-x86_64-3.6
  creating build/lib.linux-x86_64-3.6/cbor2
  copying cbor2/__init__.py -> build/lib.linux-x86_64-3.6/cbor2
  copying cbor2/types.py -> build/lib.linux-x86_64-3.6/cbor2
  copying cbor2/decoder.py -> build/lib.linux-x86_64-3.6/cbor2
  copying cbor2/compat.py -> build/lib.linux-x86_64-3.6/cbor2
  copying cbor2/encoder.py -> build/lib.linux-x86_64-3.6/cbor2
  running build_ext
  building '_cbor2' extension
  creating build/temp.linux-x86_64-3.6
  creating build/temp.linux-x86_64-3.6/source
  gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/home/circleci/project/venv/include -I/usr/local/include/python3.6m -c source/module.c -o build/temp.linux-x86_64-3.6/source/module.o
  gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/home/circleci/project/venv/include -I/usr/local/include/python3.6m -c source/encoder.c -o build/temp.linux-x86_64-3.6/source/encoder.o
  source/encoder.c: In function ‘CBOREncoder_init’:
  source/encoder.c:162:36: warning: initialization from incompatible pointer type
           PyObject *datetime_class = PyDateTimeAPI->DateType;
                                      ^
  source/encoder.c: In function ‘encode_decimal_digits’:
[...]

PR #68, opened 2020-01-17
Fix build for older gcc defaults and glibc/musl

Sekenre:
This is in response to #67 

The issue is caused by the `_cbor2` c-library using `-std=c99` but some versions of gcc default to c89.

Also in versions of glibc older than 2.9 and some versions of musl libc (not sure which ones yet) some of the 16 bit byteswap functions are missing from `endian.h`

Building the C library is temporarily disabled until these functions can be swapped out with their POSIX equivalents.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,9 @@ jobs:
 
 install: pip install tox
 
-script: tox -vv
+script:
+  - python -c "from __future__ import print_function; import platform; print(platform.libc_ver())"
+  - tox -vv
 
 after_success:
   - pip install coveralls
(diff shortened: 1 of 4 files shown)


Sekenre:
Coverage decreased because the C module was not built for any of the travis build environments. I think it depends on the `platform.libc_ver()` output not being what I expect. Expect more pushes to this PR

Sekenre:
It appears that the glibc version reported by python 3.5, 3.6 and 3.7 on travis's build farm are all less than 2.9.

This is causing coveralls to complain.

The host system Ubuntu Xenial 16.04 LTS should have a glibc of 2.23 or greater, so it seems that the python binaries were not linked to this version?

PR #69, opened 2020-01-21
Fix deprecation warning

mariano54:
This fixes the following warning:

.venv/lib/python3.7/site-packages/cbor2/decoder.py:3: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    from collections import Mapping


Sekenre:
Thanks!

Issue #71, opened 2020-02-13
CBORSimpleValue differs between C lib and Python


Sekenre:
A CBORSimpleValue returned by the C library is a subclass of namedtuple instead of the object defined in `cbor2/types.py`

I have a fix on the way.

Sekenre:
Oh yes the reason this is a problem, you can't decode a simple value used as a map key if you're using the pure python module.

PR #73, opened 2020-02-19
Fix for #71 CBORSimpleValue object differs between implementations

Sekenre:


cbor2/types.py
```diff
@@ -62,36 +63,30 @@ def __repr__(self):
         return 'CBORTag({self.tag}, {self.value!r})'.format(self=self)
 
 
-class CBORSimpleValue(object):
+class CBORSimpleValue(namedtuple('CBORSimpleValue', ['value'])):
```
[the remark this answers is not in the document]
Sekenre: the C module uses an instance of [PyStructSequence](https://docs.python.org/3/c-api/tuple.html#struct-sequence-objects) to represent simple values. It also made the implementation of  `__hash__()`, `__eq__()` and `__lt__()` shorter.

cbor2/types.py
```diff
@@ -62,36 +64,30 @@ def __repr__(self):
...
     """
     Represents a CBOR "simple value".
 
     :param int value: the value (0-255)
     """
 
-    __slots__ = 'value'
+    __slots__ = ()
+    __hash__ = namedtuple.__hash__
```
agronholm: Is this really necessary? The class already inherits from namedtuple so wouldn't that implementation of `__hash__()` be used anyway?

cbor2/types.py
```diff
@@ -62,36 +64,30 @@ def __repr__(self):
...
     def __eq__(self, other):
-        if isinstance(other, CBORSimpleValue):
-            return self.value == other.value
-        elif isinstance(other, int):
+        if isinstance(other, int):
             return self.value == other
-        return NotImplemented
+        return super(CBORSimpleValue, self).__eq__(other)
 
     def __ne__(self, other):
```
agronholm: Is this really required? Strikes me as somewhat odd to have `__eq__()` and `__ne__()` both implemented but not `__lt__ ()`. Have you that these operations require these specific implementations? If you haven't already, I suggest you check out `@total_ordering`. I don't know how much the superclass already provides, but these are the points that I wanted to bring up.

cbor2/types.py
Sekenre: Surprisingly, yes if you redefine `__eq__` on a namedtuple subclass you have to.

cbor2/types.py
Sekenre: I will look into `@total_ordering` thanks

Sekenre: Looks like I have to implement each comparison operator since `@total_ordering` operator re-orders the objects. If we want to keep the comparison to integers.

PR #78, opened 2020-04-03
Add optional flag to setuptools extension

chiefnoah:
This should prevent the whole install from failing in the event compilation fails. Especially helpful for when trying to install on Windows where Visual Studio isn't installed.

See [setuptools documentation](https://docs.python.org/3/distutils/apiref.html#distutils.core.Extension) for details.


Sekenre:
Going to try merging from the mobile app.

PR #79, opened 2020-04-07
Update usage.rst

merlinsenger:
I could not find any difference in setting datetime_as_timestamp to either true or false. I assume that the intended option is datetime_as_timestamp for this example.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -49,7 +49,7 @@ disabled by default in the next version. The behaviour can be re-enabled as foll
     from datetime import date, timezone
 
     # Serialize dates as datetimes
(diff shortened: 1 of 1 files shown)


Sekenre:
This flag is actually `date_as_datetime`  and it enables some broken behaviour that we used to have where a Python date object would be converted to a datetime prior to encoding. CBOR does not yet have a pure date datatype. If you need to send dates you need to account for this in the decoding end.

Perhaps the documentation is unclear.

merlinsenger:
My bad. I misinterpreted that for a way to use unix timestamp style instead of string style. Thanks for the quick response.

PR #80, opened 2020-04-14
Add pyproject.toml to allow PEP517 builds. `pip wheel .` now works.

richardkiss:
Make wheel building work.


agronholm:
This PR looks like a mishmash of changes here and there. I know it's marked draft, but what is your intent here, other than the one stated in the subject?

richardkiss:
It was originally just the first change. Since then, others with access to the repo have made other patches.

hoffmang9:
I did not realize that this was an open PR as well as a fork. This should probably be closed as you've addressed this another way already.

PR #83, opened 2020-05-31
Support self-described CBOR tag

kalcutter:
You are welcome to bikeshed the naming :-).

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -476,6 +476,10 @@ class CBORDecoder(object):
                     break
         raise CBORDecodeValueError("invalid ipnetwork value %r" % net_map)
 
+    def decode_self_describe_cbor(self):
+        # Semantic tag 55799
+        return self._decode()
+
     #
     # Special decoders (major tag 7)
     #
@@ -523,21 +527,22 @@ special_decoders = {
 }
 
 semantic_decoders = {
-    0:   CBORDecoder.decode_datetime_string,
-    1:   CBORDecoder.decode_epoch_datetime,
-    2:   CBORDecoder.decode_positive_bignum,
-    3:   CBORDecoder.decode_negative_bignum,
-    4:   CBORDecoder.decode_fraction,
-    5:   CBORDecoder.decode_bigfloat,
-    28:  CBORDecoder.decode_shareable,
-    29:  CBORDecoder.decode_sharedref,
-    30:  CBORDecoder.decode_rational,
-    35:  CBORDecoder.decode_regexp,
-    36:  CBORDecoder.decode_mime,
-    37:  CBORDecoder.decode_uuid,
-    258: CBORDecoder.decode_set,
-    260: CBORDecoder.decode_ipaddress,
-    261: CBORDecoder.decode_ipnetwork,
+    0:     CBORDecoder.decode_datetime_string,
+    1:     CBORDecoder.decode_epoch_datetime,
+    2:     CBORDecoder.decode_positive_bignum,
+    3:     CBORDecoder.decode_negative_bignum,
+    4:     CBORDecoder.decode_fraction,
+    5:     CBORDecoder.decode_bigfloat,
+    28:    CBORDecoder.decode_shareable,
+    29:    CBORDecoder.decode_sharedref,
+    30:    CBORDecoder.decode_rational,
+    35:    CBORDecoder.decode_regexp,
+    36:    CBORDecoder.decode_mime,
+    37:    CBORDecoder.decode_uuid,
+    258:   CBORDecoder.decode_set,
+    260:   CBORDecoder.decode_ipaddress,
+    261:   CBORDecoder.decode_ipnetwork,
+    55799: CBORDecoder.decode_self_describe_cbor,
 }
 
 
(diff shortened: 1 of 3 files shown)


Sekenre:
Hi, thanks for the contribution.

I'm ok with the tag being discarded by the decoder if it's found at the beginning of the stream. But if it found the tag part way through it would affect round trip encoding. Think about the result of `dump(outfile, load(infile))` would it have the same semantics?

On the encoding side I would prefer that this is added as a boolean option. e.g. 

```
with open('f.cbor', 'wb') as f:
    cbor2.dump(f, 1000, write_magic=True)
```

It also needs to work with the C module.

kalcutter:
I don't understand your comment about `dump(outfile, load(infile))` having the same semantics. By definition, this tag does not change the semantics of the CBOR stream. So `dump(outfile, load(infile))` gives you the same semantic stream in either case (albeit without the tag). `dump(outfile, load(infile))` has always been a lossy operation in this regard. For example, I can encode "0" as a positive bignum and this information will be lost when I do `dump(outfile, load(infile))`.

kalcutter:
There are even more egregious cases of this: If I have the array `[2, 4]` tagged as a rational, `dump(outfile, load(infile))` won't even give the same array back! It will encode the rational as `[1, 2]`. So this is truly a lossy operation even beyond the semantic tags themselves. Fixing this would require a massive rewrite of the library.

Sekenre:
Hi @kalcutter your last example seems a bit strange, roundtrips work fine for fractions:
```python
>>> import cbor2
>>> cbor2.dumps(cbor2.loads(b'\xd8\x1e\x82\x03\x02'))
b'\xd8\x1e\x82\x03\x02'
>>> cbor2.loads(b'\xd8\x1e\x82\x03\x02')
Fraction(3, 2)
```
You are correct that a bignum zero will be re-encoded as an integer, that has been debated on the IETF mailing lists, I believe the outcome was that a decoder should never expect a larger integer representation than necessary.

Self describe CBOR tag is intended to function as a file magic identifier so if you are including a CBOR file embedded in another CBOR data structure, If you discarded that tag you've modified the resulting file. It makes sense to discard the tag only if it's the first thing in your file and you've specifically requested the decoder to do so.

kalcutter:
Round trips do not work for fractions in general since Fractions are automatically simplified (from my example above):
```
>>> cbor2.dumps(cbor2.loads(b'\xd8\x1e\x82\x02\x04'))                                                                                                                                
b'\xd8\x1e\x82\x01\x02'
```
Another type type that cannot be round trip encoded in general is Datetime:
```
>>> cbor2.dumps(cbor2.loads(b'\xc0\x78\x202018-08-02T07:00:59.000251+00:00'))                                                                                                       
b'\xc0x\x1b2018-08-02T07:00:59.000251Z
```
There are many examples where round trip encoding is not possible with this library. I stand by my previous assertion that cbor2.dumps(cbor2.loads()) is lossy that no one should expect otherwise with this library.

If you are embedding CBOR in another data structure and you want it to be correctly round tripped then it needs to be stored as a byte string (preferably tagged with semantic tag 24 (CBOR data item)).

Given all of these facts, I don't see the benefit of only handling this particular semantic tag only at the start of the file. The current implementation is more consistent with how the library treats other semantic tags.

Sekenre:
Coming back to this after a rush of work, You are of course correct about the lossy nature of encoding. My apologies for being dense.

Why not use the tag hook mechanism for this? e.g.:

```python
import cbor2
from binascii import unhexlify

def drop_self_describe(decoder, tag):
    if tag.tag == 55799:
        return tag.value
    return tag

data = cbor2.loads(unhexlify('D9D9F78180'), tag_hook=drop_self_describe)
assert data == [[]]
```

Anything you want to write to a file with that tag can simply be wrapped with a cbor2.CBORTag object. e.g.

```python
import cbor2
from binascii import hexlify

myvalue = [1, 2]
encoded = cbor2.dumps(cbor2.CBORTag(55799, myvalue))
assert hexlify(encoded) == b'd9d9f7820102'
```

kalcutter:
I was using a hook before I opened the PR, however, I believe this change is a quality of life improvement for everyone using the library. I also think it fits well with how the library already handles tags. What issues do you see with merging this PR given the current lossy nature of the tag handling?

Related to this, what do you think about adding a flag to disable all built-in tag handling? This would provide users raw access to all the tags and lossless round-tripping. Certainly, if we had this feature, there would be no issue accepting this PR?

kalcutter:
I implemented the C module part.

Sekenre:
> I was using a hook before I opened the PR, however, I believe this change is a quality of life improvement for everyone using the library. I also think it fits well with how the library already handles tags. What issues do you see with merging this PR given the current lossy nature of the tag handling?

I agree it is an improvement, you've won me over! The only issue I see right now is usability of the encoder. Does the user have to remember to call `encode_self_described_cbor()` before they write out their data? Should we default to writing this tag every time we write a file?

> Related to this, what do you think about adding a flag to disable all built-in tag handling? This would provide users raw access to all the tags and lossless round-tripping. Certainly, if we had this feature, there would be no issue accepting this PR?

Yes I like that, disabling all tag handling could be very useful for protocol implementers.

kalcutter:
> Does the user have to remember to call `encode_self_describe_cbor()` before they write out their data? Should we default to writing this tag every time we write a file?

You bring up a good point about usability of encoding the tag. In 99% of cases you just want to write the tag to the beginning of the stream. Therefore, I guess we should also have a `self_describe_cbor` param to make it easier to use like: `dumps(...,self_describe_cbor=True)`.  My guess is that it should default to `False` and let the user opt-in. Although defaulting to `True` would improve adoption and support for the tag 😁. A lot of other libraries support the tag and a lot of them don't.

kalcutter:
I have updated the PR to only include the decode part since it is completely independent of encoding. I can make a separate issue to discuss the API design of encoding this tag.

PR #84, opened 2020-05-31
Fix secfrac parsing decoding semantic date/time strings

kalcutter:


kalcutter:
I updated the PR to also fix the C-extension.

kalcutter:
I added more tests. I think this PR is ready.

Sekenre:
Hi @kalcutter your tests should pass if you add the python2 unicode string prefix `u''` to your expected output.

I'm happy to merge without the s390 tests working since that seems to be a Travis issue

Sekenre:
The problem is with the string embedded in the exception.

https://github.com/agronholm/cbor2/blob/7bbaacc86b7ec4f7c8d41226a6ed04cffc40883d/cbor2/decoder.py#L380

The test could simply look for this on python2.7

```python
if sys.version_info < (3,):
    expected = "invalid datetime string: u'0000-123-01'"
else:
    expected = "invalid datetime string: '0000-123-01'"
```

kalcutter:
Oops I read the error wrong. Kind of ugly to have to deal with this everywhere like that. Would it be OK to change the exception message for python2 to be the same as python3 (without the embedded "u")?

Sekenre:
Remember, this has to work for any datatype that may be incorrectly tagged.

 Maybe it would be worth removing the diagnostic data altogether? Or adding a check to see if it's a unicode object.

kalcutter:
Fixed as per suggested above.

Issue #86, opened 2020-06-13
Installation requires a compiler


virtuald:
Currently, installing cbor2 fails if a compiler isn't installed. See https://github.com/agronholm/cbor2/blob/5e8491fe9449cbb4990d36e1fee2c93e9a10a499/setup.py#L24

There is a pure python fallback, but it seems the current setup.py will not actually use it most of the time. It would be great if the setup.py could check to see if compilation succeeds/could succeed, and fallback to pure python if the compilation fails. 

Alternatively publishing wheels for all common platforms would be great as well.

agronholm:
I was able to install this without a compiler:
`pip3 install git+https://github.com/agronholm/cbor2.git`
The latest master has the `optional=True` flag set, but the latest release does not, hence the failure.
@Sekenre time to cut a patch release?

Sekenre:
Done,
5.1.1 released with build improvements.

PR #87, opened 2020-06-14
add option to not build c-extension

jschlyter:
updates #86

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,3 +19,5 @@ wraps the ``_pickle`` C module in the Python Standard Library. On Windows, this 
 On PyPy, cbor2 runs with almost identical performance to the C backend.
 
 .. _RFC 7049: https://tools.ietf.org/html/rfc7049
+
+To build without the optional C-extension, set OS env ``CBOR2_NO_EXT``.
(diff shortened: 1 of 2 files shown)


agronholm:
There has to be a better way to make this optional.

agronholm:
Curious that it's already set as optional (see [here](https://github.com/agronholm/cbor2/blob/master/setup.py#L37)). So which tool is failing to respect that?

jschlyter:
> Curious that it's already set as optional (see [here](https://github.com/agronholm/cbor2/blob/master/setup.py#L37)). So which tool is failing to respect that?

Python 3.7 and 3.8 (from python.org and from homebrew) on MacOS builds extensions by default. I've found no other other way to disable building the extension.

Sekenre:
There is a way to do this with pip or setuptools:

```
CC="/bin/true" pip install cbor2
```

This will skip building the c extension even if you have a compiler and compatible libc.

Sekenre:
A cross-platform way to do this is to create a file called `pydistutils.cfg` in `$HOME` on unix or `%USERPROFILE%` on windows with this content:

```
[build_ext]
dry_run = 1
```

jameshilliard:
I have an improved version of this in #92 which allows for fully overriding the auto-detection behavior.

jschlyter:
Replaced by #92

PR #92, opened 2020-08-01
Allow overriding c extension build autodetection with env variable.

jameshilliard:
Similar to #87 but allows full overriding of the default behavior, useful in certain scenarios where auto-detection may be unreliable such as cross compilation.


Sekenre:
I like this. we can also set up a tox environment like so:

```ini
[testenv:py38-clib]
setenv =
    CBOR2_BUILD_C_EXTENSION = 1

[testenv:py38-noclib]
setenv =
    CBOR2_BUILD_C_EXTENSION = 0
```

jameshilliard:
@Sekenre should that be done in a separate pull request or this one?

Sekenre:
@jameshilliard in this one please, it would also make sense to update the travis.yml to run tox with these options.

PR #97, opened 2020-09-29
Switched from Travis CI to Github Actions

agronholm:
This replaces the Travis CI configuration with Github Actions workflows which test the master branch and all pull requests.
In addition, they also automatically build and upload binary wheels and a source distribution to PyPI when you push a tag matching a specific pattern.


Sekenre:
@agronholm Could you change the thresholds on coveralls to allow -1% or -2% change in coverage to pass? It generates lots of false negatives particularly when deleting code.

agronholm:
Sure. In this particular case though, it's detecting 0% coverage which is nuts, and I don't know why.

agronholm:
I've switched the coverage decrease threshold to 3%.

agronholm:
After running the test suite: `Coverage.py warning: No data was collected. (no-data-collected)`. This is despite a detailed coverage chart being printed which indicates 90% coverage. WTF?

agronholm:
Ok, that takes care of the coverage problem. The reason coverage fell that much was because I did not include Python 2.7 in the test array. I'll add that too I suppose.

agronholm:
I fixed the 2.7 test errors but the highest coveralls version supported on 2.7 is so old it won't work with Github Actions. Bummer. How do you want to handle this?

Sekenre:
I think tag the current master and update documentation to say this will be the last version tested on 2.7, then merge this PR. How does that sound?

agronholm:
Sounds good. You may want to drop support for all Pythons older than 3.6 at the same time since 3.5 has passed its EOL date.

agronholm:
Note that you should really drop 2.x compatibility code at the same time; not doing that **will** lead to 2.x breakage. I can make a PR for that if you want.

Sekenre:
If you have time that would be awesome!

agronholm:
I can probably do that for the Python code but not for the C code as I didn't write any of that.

waveform80:
Apologies I've been absent for a while! I can try and take a look at integrating coverage for the C bits if wanted; it looks like coveralls has (community contributed) [support for gcov](https://github.com/eddyxu/cpp-coveralls) output which is what my little [coverage script](https://github.com/agronholm/cbor2/blob/master/scripts/coverage.sh) used (lcov is just a gcov front-end).

I think there's still some 3.5 compatibility bits in the C code too which I can excise at the same time.

agronholm:
I guess it's ready to be merged now. I will follow up with a PR that removes other py2.7 specifics.

PR #98, opened 2020-09-30
Removed support for Python < 3.6

agronholm:
This also adds test jobs and markers for Python 3.9.


agronholm:
This PR is built on top of the one that adds the Github Actions workflow.

If you wish to retain Python 3.5 support (even though it's reached its EOL), I can make the necessary modifications.

Sekenre:
Thanks Alex, I would prefer not to support Python 3.5.

PR #100, opened 2020-11-10
New sdist, handle remote PRs/forks

hoffmang9:
- pep517 build is being deprecated
- By checking for secrets access PRs can run against these and build cleanly - not as critical for this action but useful other places

Is there a reason you haven't published 2.5.0 binary wheels to pypi?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -53,9 +53,9 @@ jobs:
       with:
         python-version: 3.x
     - name: Install dependencies
-      run: pip install pep517
+      run: pip install build
     - name: Create sdist
-      run: python -m pep517.build --source .
+      run: python -m build --sdist --outdir dist .
     - uses: actions/upload-artifact@v2
       with:
         name: sdist
@@ -74,7 +74,17 @@ jobs:
       run: |
         mkdir dist
         mv */*.whl */*.tar.gz dist
+    - name: Test for secrets access
+      id: check_secrets
+      shell: bash
+      run: |
+        unset HAS_SECRET
+        if [ -n "$SECRET" ]; then HAS_SECRET='true' ; fi
+        echo ::set-output name=HAS_SECRET::${HAS_SECRET}
+      env:
+        SECRET: "${{ secrets.secrets.pypi_password }}"
     - name: Upload packages
+      if: steps.check_secrets.outputs.HAS_SECRET
       uses: pypa/gh-action-pypi-publish@master
       with:
         user: __token__


agronholm:
The machinery for automated wheel builds was merged after the 2.5.0 release IIRC.
I didn't know about [build](https://pypi.org/project/build/). Thanks for the tip.

hoffmang9:
Great - apologies for the earlier PR spam. We were keeping a fork to build binary wheels for cbor2. We'll still need to do that for ARM64, but that's a far simpler proposition now that you'll be publishing binary wheels to pypi.

agronholm:
Why is there a step for checking secrets access?

hoffmang9:
If someone forks and they don't have that secret, then instead of the ci failing, it will succeed but skip the upload to pypi (or test pypi if you add that) step cleanly. Also handles that in a pull request.

agronholm:
The publish workflow will only be triggered when a specific kind of tag is pushed. Have you still run into problems with this?

hoffmang9:
Granted that we're an edge case, but I had the ci fail in my fork when I tagged it 2.5.0 to build wheels of 2.5.0.

This is a general nice to have when using GitHub Actions for forks (and PR's should you start building on each PR to e.g. test pypi.)

agronholm:
This has the downside that if the secret is missing when it's really supposed to upload to PyPI, the job will not error out, right?

hoffmang9:
Yes but if the secret is missing the upload to Pypi is doomed to fail anyway.

hoffmang9:
However, in that case you would at least get your artifacts that you could manually submit to Pypi. I don't think you get the artifacts in the failed Action scenario.

agronholm:
The workflow can be manually restarted in such a case.

hoffmang9:
That's true wether you fail quietly or fail. Feel free to excise the secrets check - it's simply an enhancement for third party contributors.

agronholm:
@Sekenre it's your call, but I'm favoring removing the secrets check.

Sekenre:
Crazy idea: You can change the URL for the publish step https://github.com/pypa/gh-action-pypi-publish to test.pypi.org If we make the url parametrizable then each person who wants to test that step can set up their own credentials on test.pypi.org?

agronholm:
Master has now been updated to use `build` instead of `pep517` and a PyPI token has been activated. The next time a new version is published, wheels will be built and uploaded. I will close this since @Sekenre has not taken any action on it.

PR #101, opened 2020-11-13
String references

xurtis:
This adds the ability to decode [string references](http://cbor.schmorp.de/stringref) to both the C and Python decoders.

I've not worked on Python & C interop before, so I may have missed some things.

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -39,7 +39,7 @@ class CBORDecoder:
 
     __slots__ = (
         '_tag_hook', '_object_hook', '_share_index', '_shareables', '_fp_read',
-        '_immutable', '_str_errors')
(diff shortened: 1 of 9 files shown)


Sekenre:
Thanks @xurtis This looks useful, do you plan to write an encoder? Without round-trip encode/decode I can't actually make use of this feature.

xurtis:
Should be able to add an encoder without too much trouble.

xurtis:
@sekenre I've updated the PR to include an encoder.

agronholm:
This looks clean – do you have any remaining doubts about this PR @Sekenre ?

Sekenre:
This is great thanks @xurtis, sorry about the wait.

Issue #105, opened 2021-03-02
SystemError raised decoding under-long bytes


chrysn:
When loading this byte string that came up in an EDHOC plug test, I get a SystemError:

```
>>> cbor2.loads(b'\xc7\x887\x00\x16\xb8\x96[\xdb t\xbf\xf8.Z \xe0\x9b\xec!\xf8@n\x86D+\x87\xec?\xf2E\xb7\nGbM\xc9\xcd\xc6\x82K*LR\xe9^\xc9\xd6\xb0SKq\xc2\xb4\x9eK\xf9\x03\x15\x00\xce\xe6\x86\x99y\xc2\x97\xbbZ\x8b8\x1e\x98\xdbqA\x08A^\\P\xdbx\x97L\'\x15y\xb0\x163\xa3\xefbq\xbe\\"^\xb2')
SystemError: <built-in function loads> returned NULL without setting an error
```

This happens both with the cbor2 shipped with Debian (5.2.0-4) as well with the current master 818ba624df535ac7cd300aaa6fb963a83b3656e0 (tested for `_cbor2.__file__` to be in the virtualenv I installed it in).

chrysn:
Maybe to clarify a bit: This came up because a library was using cbor2 for heuristics on whether something looks liike it is CBOR or not. That's probably a bad thing to do (especially here where the data is not supposed to be CBOR), but then again shouldn't raise this kind of exception either.

Sekenre:
When tested with the python impementation this gives an `OverflowError` exception. We need a way to check if a decoded integer can be used as a length (PY_SSIZE_T_MAX)

chrysn:
Thank you!

Issue #99, opened 2020-10-11
Test failures on i386


bgermann:
Running the test suite on i386 an error triggers: cannot fit 'int' into an index-sized integer
See https://salsa.debian.org/python-team/packages/cbor2/-/jobs/1064086

lopsided98:
I'm seeing this on armv7l as well.

Sekenre:
These particular tests are supposed to fail, but because it's a 32bit platform it's generating different exceptions. I will try to replicate this on my own hardware once I have time.

The main thing we want to avoid is the `_cbor2` module trying to pre-allocate all of the platform's RAM in the event of malformed data.

lopsided98:
What I believe is happening is that the encoded length of the string doesn't fit in a 32-bit integer, so it fails with the overflow error before getting to the error the test is supposed to catch.

PR #113, opened 2021-05-28
Bounds checks on decoding

Sekenre:
Intended as a fix for #105 and #99, may improve our security story if done properly.

Includes tests for the most obvious problems.

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -1,5 +1,6 @@
 import re
 import struct
+import sys
 from collections.abc import Mapping
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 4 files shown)


Sekenre:
You might think this all works, but it's badly broken on 32bit i386 and probably 32bit ARM. Unfortunately I can't see a way to test 32bit builds on Github

Sekenre:
Note, we don't need bounds checks on decoding maps in C since we don't try to pre-allocate anything. Any decode failures are picked up as we decode each key-value pair and cPython handles the allocation. Also the map length isn't passed on to the c-api at any point.

PR #115, opened 2021-06-24
Fix #114: Raise TypeError when using a negative tag

### LINKED ISSUE #114: Negative CBORTag values are encoding without error in the C encoder
Example:

```python
>>> # Pure python module
>>> from cbor2.encoder import dumps
>>> from cbor2.types import CBORTag
>>> dumps(CBORTag(-1, 'val'))
  ...
struct.error: ubyte format requires 0 <= number <= 255

>>> # c module
>>> import cbor2
>>> cbor2.dumps(cbor2.CBORTag(-1, 'val'))
b'\xdb\xff\xff\xff\xff\xff\xff\xff\xffcval'
```

CBORTag should raise a value error with a negative int

Sekenre:
TODO: Fix c-module

### DIFF
--- a/cbor2/types.py
+++ b/cbor2/types.py
@@ -44,8 +44,8 @@ class CBORTag:
     __slots__ = 'tag', 'value'
 
     def __init__(self, tag, value):
-        if not isinstance(tag, int):
-            raise TypeError('CBORTag tags must be integer numbers')
+        if not isinstance(tag, int) or tag not in range(2**64):
+            raise TypeError('CBORTag tags must be positive integers less than 2**64')
         self.tag = tag
         self.value = value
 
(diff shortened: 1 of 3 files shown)


Sekenre:
All done. Note that this does not quite conform to spec here: https://www.ietf.org/archive/id/draft-bormann-cbor-notable-tags-03.html#section-toc.1-1.10.1

It's not clear to me why you would need to specifically consider a tag as invalid.

Sekenre:
Got some feedback here: https://mailarchive.ietf.org/arch/msg/cbor/m8z-7iOK8ApXyfZVVBcb-gsY2Ww/

Basically an encoder should never emit these tags, but the decoder could use them to mark invalid values instead of raising an exception.

PR #126, opened 2022-01-30
Build arm64 wheels for Linux, macOS

agronholm:
This also drops the building of musllinux wheels as they add a lot to the total build time. Your call though.


agronholm:
Do we want to build 32-bit Windows wheels too?

Sekenre:
We seem to have a lot of users that want to build Alpine Linux based Docker images, the musl wheel build is for their benefit. How much time would it add?

I'm ok with removing it as long as we include instructions for those users.

agronholm:
If that is the case then the extra time building musl wheels would be justified. I'll add it back.

agronholm:
All good now?

PR #132, opened 2022-04-26
Migrated packaging metadata to pyproject.toml

agronholm:
This also bumps the minimum setuptools and setuptools_scm versions, requires to handle project metadata being in pyproject.toml.
The "Pure Python" part from the project description was also removed since it has had a C extension for a while.


agronholm:
It would seem that this requires some shuffling of the code files, plus dropping of Python 3.6 support.

agronholm:
We've never had a "Homepage" URL, and the packaging tools are not smart enough to determine that our 3 project URLs are a good enough substitute.

Sekenre:
It's the future deprecation warning I'm more worried about. Is that definitely the way to go?

agronholm:
I'm pretty confident that they won't remove support. This I say as a PyPA member who has been participating in some of the discussions around this topic.

Sekenre (approved):
Looks good!

agronholm:
By the way, you should never, ever do `python setup.py install` anymore. This is a thing of the past, obsoleted by pip's support for [PEP 517](https://peps.python.org/pep-0517/).

agronholm:
Thanks for the review!

PR #135, opened 2022-05-03
Bumped up the Python requirement to 3.7+

agronholm:
My previous PR missed the change to bump the minimum Python requirement to 3.7. This also removes the 3.6 classifier.

When this is merged, you can force-add the 5.4.3 tag and force-push it to GitHub. That will trigger the publish process anew.


agronholm:
Do you intend to update the tag? I can do that too if you like.

Sekenre:
Yes sure I'm away from my laptop at the moment.

PR #143, opened 2022-10-26
decoder.py: Raise CBORDecodeValueError on invalid bignum values

oyvindronningstad:
This matches the behavior of the C code. Code that expects a CBORDecodeValueError on invalid data will fail without this patch when running without the C extension, since the current code raises a ValueError.

I discovered this when I installed cbor2 on Python 3.11.0rc1, since it installed cbor2 without C extensions.

### DIFF
--- a/cbor2/decoder.py
+++ b/cbor2/decoder.py
@@ -454,6 +454,8 @@ class CBORDecoder:
         from binascii import hexlify
 
         value = self._decode()
+        if not isinstance(value, bytes):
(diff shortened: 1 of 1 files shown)


agronholm:
If you merge master now, it should fix the pre-commit failure. We will also need to test against Python 3.11, but I will do that in a PR of my own.

oyvindronningstad:
@agronholm Thanks, rebased

Issue #147, opened 2022-11-25
TypeError: unhashable type: '_cbor2.CBORTag'


nielstron:
The following code gives me a rather obscure error message. `TypeError: unhashable type: '_cbor2.CBORTag'`
```
import cbor2

cbor = bytes.fromhex("d8799f4469555344d8799fd8799f1b0dd612c92160462cffd8799fc24d0af6e24f990db06e7dd5a6a810ffd8799f1b008b36bf518dee53ff0100ffa2d8799f0000ffd8799f1b084494e2d23b2b7effd8799f0100ffd8799f1b008b36bf518dee53ffff")
res = cbor2.loads(cbor)
print(res)
```
However, using cbor.me, I am perfectly able to correctly load the cbor. I assume there is a bug in the implementation?

nielstron:
Interesting enough, the pure command line usage throws no error:
`echo d8799f4469555344d8799fd8799f1b0dd612c92160462cffd8799fc24d0af6e24f990db06e7dd5a6a810ffd8799f1b008b36bf518dee53ff0100ffa2d8799f0000ffd8799f1b084494e2d23b2b7effd8799f0100ffd8799f1b008b36bf518dee53ffff | xxd -r -ps | python -m cbor2.tool --pretty`

Sekenre:
Hi @nielstron Thanks for raising this one. At the moment, we're trying to keep functionality identical between the C library `_cbor2` and the pure python implementation. Since we don't have a `__hash__` method on the C object, we can't have one on the Python object.

However, a change I intend to make for the next major version update is to always use the pure python tag object. We would then be free to make it hashable.

It works in the command line because the default tag_handler in that module converts tags to their string equivalent (purely diagnostic, not reversible)

You can also work around this by adding a tag handler that can return a hashable object of your choosing.

nielstron:
If that is the issue, I just pushed a version in #148 that adds an equivalent hash function to both the python version and the C version of the object.

Sekenre:
Merged. That's been bugging me for ages and I had been putting off learning how to do it in Python's C API

PR #148, opened 2022-11-25
Make CBORTag hashable

nielstron:
This is a trivial solution to #147 by making the CBORTag hashable.

### DIFF
--- a/cbor2/types.py
+++ b/cbor2/types.py
@@ -63,6 +63,9 @@ class CBORTag:
     def __repr__(self):
         return "CBORTag({self.tag}, {self.value!r})".format(self=self)
 
+    def __hash__(self):
+        return hash((self.tag, self.value))
+
 
 class CBORSimpleValue(namedtuple("CBORSimpleValue", ["value"])):
     """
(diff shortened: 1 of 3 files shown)


nielstron:
There appears to be a test case that specifically checks that CBOR tags are not hashable - why is that the case? Because the class is not immutable? Why does this error have to be thrown when trying to *load* an object? The fact that it is not hashable should IMO not completely forbid it to be created at all or am I missing something?

The testcase: https://github.com/agronholm/cbor2/blob/aef7e88d3ef2bb6985ce4d8854f4a5ad5e860c04/tests/test_decoder.py#L701

agronholm:
Did the test suite pass for you locally?

agronholm:
@waveform80 could you comment? Why does `CBORTag` have to be unhashable?

Sekenre:
That's brilliant! Thank you

nielstron:
Follow up: Will there be a new minor version of the package with this fix included?

agronholm:
Given that there are other changes (the no-arch wheel) that went into `master` recently, that would be nice. @Sekenre ?

Sekenre:
Yup, coming out later today.

PR #149, opened 2022-11-26
Updated GitHub actions

### LINKED ISSUE #146: Provide -none-any wheels
For applications in which there is neither access to a compiler nor there is any widespread CPU architecture used, it would be convenient to provide -none-any wheels in addition to the provided binary wheels.

A concrete use case for this is pyodide, which runs Python code in the browser: At <https://pyodide.org/en/stable/console.html>, running `import micropip; await micropip.install('cbor2'), micropip.list()` shows that it falls back to cbor2 version 4.1.2, which was the last version that had such wheels.

One reason not to do this would be if pip preferred these to the platform specific ones; I don't think it does, but I did not try it.

An alternative for that particular use case might be to ask pyodide to provide the module within their build system, which would compile the C code to WASM ([they managed even with cryptography](https://github.com/pyodide/pyodide/issues/761)).

agronholm:
* Publish a pure-Python wheel
* Updated the cibuildwheel, gh-action-pypi-publish and download-artifact actions to their latest versions
* Fixed Python cache not being used on macOS and Windows
* Removed useless setup-python and cache steps in the binary wheel build job
* Removed redundant `user` option in the publish action

Closes #146.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,6 @@ name: Publish packages to PyPI
 
 on:
   push:
-    branches: [update-actions]
     tags:
       - "[0-9]+.[0-9]+.[0-9]+"
       - "[0-9]+.[0-9]+.[0-9]+.post[0-9]+"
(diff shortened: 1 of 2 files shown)


Sekenre (approved):
Looks better and more comprehensive than my solution. Thanks!

PR #150, opened 2022-11-28
Fix potential memory leak in cbortag hash function

nielstron:
I fear I have introduced a memory leak with the implementation of the hash function. I have not been able to detect it properly, but digging through the documentation of the C-Python API gives that at least
- the reference count of the object passed into the tuple is increased by one when invoking Py_BuildValue
- probably the callee is in charge of handling the decrease of the reference count of the object created in Py_BuildValue

This also properly handles the case where the tuple allocation might fail.

### DIFF
--- a/scripts/ref_leak_test.py
+++ b/scripts/ref_leak_test.py
@@ -21,8 +21,6 @@ from fractions import Fraction
 
 import objgraph
 
-import cbor2
-
 
 def import_cbor2():
     # Similar hack to that used in tests/conftest to get separate C and Python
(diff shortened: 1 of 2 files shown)


Sekenre:
Is it serious enough to warrant pulling the release? At the moment there's a failing test when trying to build a `CBORTag` object

nielstron:
Okay with these error messages I could precisely find the issue. So currently during each call to CBORTag.hash a tuple (CBORTag.tag, pointerToCBORTag_value) is created and leaked, preventing the value in the CBORTag to be deallocated as well. This PR should entirely fix this.

Issue #163, opened 2023-06-02
Python decoder implementation ignores timezone sign


rudigiesler:
This was picked up through the hypothesis testing.

Due to the way this is implemented in the python version with the regex, the sign and the hour of the timezone are combined into an integer. This causes issues where the hour is zero, but the timezones offset is negative (eg. `-00:30`), as the sign gets lost in the conversion to integer.

Example:
```python
>>> module.loads(b'\xc0x\x192000-01-01T00:00:00-00:01')
datetime.datetime(2000, 1, 1, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=60)))
```

PR #164, opened 2023-06-02
Remove regex parsing of timestamps

rudigiesler:
This is from the DjangoCon EU sprints.

`fromisoformat` was added in Python 3.7.

Since Python 3.6 support has been dropped, we can now use it instead of the previous regex implementation.

This should also fix https://github.com/agronholm/cbor2/issues/163

### DIFF
--- a/tests/test_decoder.py
+++ b/tests/test_decoder.py
@@ -12,7 +12,6 @@ from ipaddress import ip_address, ip_network
 from uuid import UUID
 
 import pytest
-
 from cbor2.types import FrozenDict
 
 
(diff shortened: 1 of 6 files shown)


tests/test_decoder.py
```diff
@@ -408,18 +407,6 @@ def test_datetime_secfrac_overflow(impl):
...
-        impl.loads(b"\xc0\x78\x1a2018-08-02T07:00:59.+00:00")
-    assert isinstance(excinfo.value, ValueError)
-    assert str(excinfo.value) == "invalid datetime string: '2018-08-02T07:00:59.+00:00'"
-
-    with pytest.raises(impl.CBORDecodeError) as excinfo:
-        impl.loads(b"\xc0\x78\x152018-08-02T07:00:59.Z")
-    assert isinstance(excinfo.value, ValueError)
-    assert str(excinfo.value) == "invalid datetime string: '2018-08-02T07:00:59.Z'"
-
-
```
rudigiesler, line 422: `fromisoformat` handles this and imports the timestamp correctly, so we do not require this test.

PR #167, opened 2023-06-02
Handle UnicodeDecodeError correctly.

### LINKED ISSUE #156: UnicodeDecodeError raised instead of some CBORDecodeError
I would have expected a `CBORDecodeValueError` to be raised and caught by the following code:

```
try:
  cbor2.loads(b'\x61\x98')
except CBORDecodeError:
  ...
```

Is this something I can/should write a PR for?

Sekenre:
This wraps the unicode error in a `CBORDecodeValueError` since a string type that is not valid utf-8 is also not valid CBOR.

Fixes #156


agronholm:
This has been addressed in a different PR.

Issue #198, opened 2023-12-20
unexpected exceptions raised while parsing untrusted inputs using cbor2.loads


robertstypa:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### cbor2 version

5.5.1

### Python version

3.10.12

### What happened?

I have a script which is parsing untrusted data using the `cbor2.loads` method.  This script is trying to verify if the provided data is cbor encoded.

The implementation was as follows:
`try:
  cbor2.loads(b'\x959;{{{{{{{{{{{{{')
except CBORDecodeError:
  print('no cbor encoded')`

For some inputs, I've noticed that `MemoryError` is raised instead of `CBORDecodeError`. 

To better understand the problem and ensure that this is only one strange case while parsing untrusted data I've run fuzzer against cbor2.loads method. 

It seems that the `cbor2.loads` method is not able to parse untrusted data properly - in the worst case cbor2 is trying to allocate the whole memory - ref to the `MemoryError' case presented in the code above.

I was able to find following exceptions raised  by cbor2.loads (all reproduced using cbor2 5.5.1/python 3.10.12/Ubuntu 20.4):
```
# OverflowError: timestamp out of range for platform time_t
cbor2.loads(b'\xc1\x1b\x9b\x9b\x9b\x00\x00\x00\x00\x00')
# OSError: OSError: [Errno 75] Value too large for defined data type
cbor2.loads(b'\xc1\x1b\x16\x16\x16\x16\x16\x16\x16\x16\x16\x16\x16\x16\x16\x16')
# MemoryError:
cbor2.loads(b'\x959;{{{{{{{{{{{{{')
# TypeError: object of type 'int' has no len()
cbor2.loads(b'\xd8%\x00\x10`\x00\x00\x00`\x10\x00\x00\x00\x00\x00\x00')
# SystemError: <built-in function loads> returned NULL without setting an error
cbor2.loads(b'\xd8\x1e\x84\xff\xff\xff\xff')
# re.error: unbalanced parenthesis at position 0
cbor2.loads(b'\xd8#A)')
# UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 1: invalid start byte
cbor2.loads(b'b\n\xff')
```

I was trying to analyze how it could be improved but it is not an easy task for somebody who does not maintain this code. Is it possible to improve it somehow?

[...]

PR #202, opened 2023-12-28
Fixed infinite recursion on recursive hash() on CBORTag

agronholm:
Relates to #198.

### DIFF
--- a/cbor2/_types.py
+++ b/cbor2/_types.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import threading
 from collections import namedtuple
 from collections.abc import Iterable, Iterator
 from functools import total_ordering
@@ -8,6 +9,8 @@ from typing import Any, Mapping, TypeVar
 
 KT = TypeVar("KT")
 VT_co = TypeVar("VT_co", covariant=True)
+
+thread_locals = threading.local()
 
 
 class CBORError(Exception):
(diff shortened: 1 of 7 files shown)


agronholm:
This was no low hanging fruit, I tell you :) It was the hardest fix to date, for me at least.

agronholm:
What input is producing segfaults now? I only remember one being mentioned, but I fixed that one already.

mschwager:
> This was no low hanging fruit, I tell you :) It was the hardest fix to date, for me at least.

Sorry, I meant low hanging fruit for the fuzzer to find crashes, not for the fix. By low hanging fruit I just meant the fuzzer was able to find the crashes in minutes or hours rather than days or weeks of runtime :+1: 

> What input is producing segfaults now? I only remember one being mentioned, but I fixed that one already.

As far as segfault crashes, there were two:

- https://github.com/agronholm/cbor2/issues/198#issuecomment-1869630196
- https://github.com/agronholm/cbor2/issues/198#issuecomment-1869730701

I re-ran the fuzzer against this branch and neither of those crashes, or any others, appeared after a few hours of runtime.

agronholm:
Ok, we're on the same page then. Perhaps the `MemoryError` still needs to be investigated, but the rest of the errors just need to be wrapped in `CBORDecodeError`.

PR #212, opened 2024-01-30
Add CIFuzz GitHub Action

mschwager:
Now that `cbor2` has been integrated into OSS-Fuzz, I thought it would be helpful to run a short CIFuzz job on PRs. This Action is pretty much verbatim from the OSS-Fuzz CI docs here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -42,7 +42,7 @@ jobs:
           path: ./out/artifacts
       - name: Upload Sarif
         if: always() && steps.build.outcome == 'success'
-        uses: github/codeql-action/upload-sarif@v3
+        uses: github/codeql-action/upload-sarif@v4
         with:
           # Path to SARIF file relative to the root of the repository
           sarif_file: cifuzz-sarif/results.sarif
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,6 +1,7 @@
 name: CIFuzz
 
 on:
+  workflow_dispatch:
   pull_request:
     # Fuzzing is currently only targeting the C extension
     paths:
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -2,7 +2,7 @@ name: CIFuzz
 
 on:
   pull_request:
-    # Fuzzing is currently only targetting the C extension
+    # Fuzzing is currently only targeting the C extension
     paths:
       - "**.c"
       - "**.cc"
@@ -41,7 +41,7 @@ jobs:
           path: ./out/artifacts
       - name: Upload Sarif
         if: always() && steps.build.outcome == 'success'
-        uses: github/codeql-action/upload-sarif@v2
+        uses: github/codeql-action/upload-sarif@v3
         with:
           # Path to SARIF file relative to the root of the repository
           sarif_file: cifuzz-sarif/results.sarif
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,48 @@
+name: CIFuzz
+
+on:
+  pull_request:
+    # Fuzzing is currently only targetting the C extension
+    paths:
+      - "**.c"
+      - "**.cc"
+      - "**.cpp"
+      - "**.cxx"
+      - "**.h"
+
+permissions: {}
+
+jobs:
+  fuzz:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+      - name: Build Fuzzers
+        id: build
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+        with:
+          oss-fuzz-project-name: "cbor2"
+          language: python
+          dry-run: true
+      - name: Run Fuzzers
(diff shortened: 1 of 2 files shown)


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,48 @@
...
+    permissions:
+      security-events: write
+    steps:
+      - name: Build Fuzzers
+        id: build
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+        with:
+          oss-fuzz-project-name: "cbor2"
+          language: python
+          dry-run: true
```
mschwager, line 27: I enabled `dry-run` for now until we confirm this Action is working as expected.

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Run Fuzzers
+        uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+        with:
+          oss-fuzz-project-name: "cbor2"
+          language: python
+          fuzz-seconds: 600
+          output-sarif: true
+          dry-run: true
```
mschwager, line 35: Same.

mschwager:
Looks like the CIFuzz job ran successfully: https://github.com/agronholm/cbor2/actions/runs/7715192673.

Let me know if you'd like to keep the `dry-run` option set, or if you'd like the CIFuzz job to fail if there are findings.

agronholm:
Should I expect the current state of the job to find anything? I *know* there are issues in the current code base that ought to be found by a fuzzer.

mschwager:
> Should I expect the current state of the job to find anything? I _know_ there are issues in the current code base that ought to be found by a fuzzer.

I ran the fuzzer locally for 10min (what CIFuzz is running for) and didn't produce any crashes. A fuzzer producing issues will ultimately depend on a number of factors such as fuzz time, input corpus, and code path selection. In this case, I would consider CIFuzz to be a quick sanity check that changes to C code aren't producing any trivial crashes. Whereas, I would consider OSS-Fuzz runs to be more comprehensive, longer running, and more likely to find crashes/issues.

One thing I did notice [in the CIFuzz run](https://github.com/agronholm/cbor2/actions/runs/7715192673/job/21029145584) was that the coverage seemed low. It's showing `cov: 26`, whereas locally I'm seeing values like `cov: 850` and higher. I'm not sure if that's because CIFuzz is doing some kind of diff-aware fuzzing where it only fuzzes the modified code (in this case a minimal comment in C code), or if CIFuzz is having trouble with the coverage instrumentation.

agronholm:
I tried to add a possibility to run this manually, but to no avail. Any idea why `workflow_dispatch:` isn't doing anything? I'd like to have the ability to run this against any branch when I want, not just PRs.

mschwager:
> I tried to add a possibility to run this manually, but to no avail. Any idea why `workflow_dispatch:` isn't doing anything? I'd like to have the ability to run this against any branch when I want, not just PRs.

Hmm, I think the workflow itself has to exist in the default branch before it can be manually dispatched. I think this is saying roughly the same thing: https://stackoverflow.com/questions/75250667/github-workflow-workflow-dispatch-missing-in-actions-tab.

In other words, the workflow has to be merged into the default branch, then it can be manually dispatched against any branch 👍 .

agronholm:
Perhaps this is causing the problem with coverage?
```
2024-02-01 22:17:15,592 - root - ERROR - Unable to download from: https://storage.googleapis.com/cbor2-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/cbor2_loads_fuzzer/public.zip. Code: 403. Content: b"<?xml version='1.0' encoding='UTF-8'?><Error><Code>AccessDenied</Code><Message>Access denied.</Message><Details>Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object. Permission 'storage.objects.get' denied on resource (or it may not exist).</Details></Error>".
2024-02-01 22:17:15,593 - root - WARNING - Failed to download corpus for loads_fuzzer.
```

agronholm:
One more thing: am I wrong, or does it always run the fuzzing on the `master` branch?

mschwager:
> One more thing: am I wrong, or does it always run the fuzzing on the `master` branch?

Hmm, I don't think so. In the [CI job for this PR](https://github.com/agronholm/cbor2/actions/runs/7715192673/job/21029145584) I see the following:

```
2024-01-30 17:38:01,735 - root - DEBUG - Stderr of command "git fetch origin refs/pull/212/merge" is: From https://github.com/agronholm/cbor2
 * branch            refs/pull/212/merge -> FETCH_HEAD
.
2024-01-30 17:38:01,746 - root - DEBUG - Stderr of command "git checkout -f FETCH_HEAD" is: Note: switching to 'FETCH_HEAD'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at c051c1e Merge 75ae7e0718472549bacb8e7b1f8c7525e1c79b12 into 8a6da705e6a85ffb8442feaeee5fd57a3d47f528
```

PR #218, opened 2024-02-18
Fixed epoch date decoding being affected by the local time zone

### LINKED ISSUE #209: Failing unit-test
### Things to check first

### cbor2 version

5.6.0

### Python version

3.11.5

### What happened?

I ran `python -m pytest -sv tests/` and one of the date tests failed:

```
tests/test_decoder.py:389: in test_date
    assert decoded == expected
E   assert datetime.date(2018, 12, 30) == datetime.date(2018, 12, 31)
```

run `python -m pytest -sv tests/test_decoder.py`

agronholm:
Fixes #209.

### DIFF
--- a/source/module.c
+++ b/source/module.c
@@ -621,6 +621,7 @@ _CBOR2_init_thread_locals(void)
 
 PyObject *_CBOR2_empty_bytes = NULL;
 PyObject *_CBOR2_empty_str = NULL;
+PyObject *_CBOR2_date_ordinal_offset = NULL;
 PyObject *_CBOR2_str_as_string = NULL;
 PyObject *_CBOR2_str_as_tuple = NULL;
 PyObject *_CBOR2_str_bit_length = NULL;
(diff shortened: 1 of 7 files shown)


source/module.c
```diff
@@ -1000,6 +1003,9 @@ PyInit__cbor2(void)
 
 #undef INTERN_STRING
 
+    if (!_CBOR2_date_ordinal_offset &&
+            !(_CBOR2_date_ordinal_offset = PyLong_FromLong(719163)))
```
meejah, line 1007: Would be helpful (to me, who doesn't know CBOR expect as a user) to know what this magic number means.

agronholm: See the CBOR definition for [semantic tag 100](https://datatracker.ietf.org/doc/html/draft-ietf-cbor-date-tag-04#section-2.1). 719163 is the number of days between the minimum supported CBOR date and the [minimum supported Python date ordinal](https://docs.python.org/3/library/datetime.html#datetime.date.fromordinal) (which starts from January 1 of year 1).

agronholm:
Turns out the call was faulty and caused the C extension to fail compilation.

PR #219, opened 2024-02-18
Changed/corrected type annotations

### LINKED ISSUE #216: Incorrect mypy types
### Things to check first

### cbor2 version

5.6.1

### Python version

3.11

### What happened?

Running mypy on my code which uses cbor2 revealed some things which appear to be incorrect types
- https://github.com/agronholm/cbor2/blob/5.6.1/cbor2/_decoder.py#L794 (should be `-> Any`)
- https://github.com/agronholm/cbor2/blob/5.6.1/cbor2/_decoder.py#L825 (should be `: Io[bytes]`)
- https://github.com/agronholm/cbor2/blob/5.6.1/cbor2/_decoder.py#L829 (should be `-> Any`)
- https://github.com/agronholm/cbor2/blob/5.6.1/cbor2/_encoder.py#L751 (should be `: Any`)
- https://github.com/agronholm/cbor2/blob/5.6.1/cbor2/_encoder.py#L752 (should be `: Io[bytes]`)

Add mypy to CI and I believe it will catch this.

agronholm:
Fixes #216.

### DIFF
--- a/cbor2/_decoder.py
+++ b/cbor2/_decoder.py
@@ -7,7 +7,7 @@ from codecs import getincrementaldecoder
 from collections.abc import Callable, Mapping, Sequence
 from datetime import date, datetime, timedelta, timezone
 from io import BytesIO
-from typing import TYPE_CHECKING, Any, TypeVar, cast, overload
+from typing import IO, TYPE_CHECKING, Any, TypeVar, cast, overload
 
 from ._types import (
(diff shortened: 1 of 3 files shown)


abrisco (approved):
This change looks good to me! It even looks more correct than what I initially reported. Thank you for updating this <3

Issue #224, opened 2024-03-13
Better support for lists


theeldermillenial:
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

The decoding mechanism doesn't have a way of differentiating between an indefinite array and a fixed length array such that a round trip decode and encode could be performed without loss of information. 

Ideally fixed length arrays are given a different class type than variable length arrays. For example, a fixed length array may be decoded as a `tuple` or `deque` while an indefinite array could be assigned the type `list`.

### Use case

In `pycardano`, cbor is hashed. If an array of length two is defined as an indefinite array (`/x9f`) rather than an array of length 2, decoding then encoding yields a different cbor result, which gives a different hash. This is problematic when verifying cbor contents.

Since cbor2 does not distinguish between fixed/indefinite arrays, pycardano creates a custom encoder that is used to create an indefinite array when requested (even if the array is smaller than 30 values). However, there is no analogous functionality with decoding, and trying to subclass CBORDecoder will not be straightforward. 

Thus, the ideal implementation would be to differentiate between these two encodings by using different classes. One mechanism could be as described, or alternatively dummy `list` classes could be created to distinguish between the two (e.g. FixedArray and IndefiniteArray, both of which are just lists).

I am happy to implement this in any way @agronholm or any other maintainer would like. Just point me in the right direction. The goal is round trip reproduction of cbor regardless of how the array is encoded.

T-recks:
Are there any updates on this issue? I need a way to encode indefinite length arrays using `cbor2.dumps()` and can't tell for sure if there's currently any way to do this.

theeldermillenial:
@T-recks I do have an open PR on this, but it's currently held up by it causing a divergence from the C extension.

https://github.com/agronholm/cbor2/pull/225

I have not had a chance to come back to it to finish it off. Basically, the request was that I modify the C-extension to make sure that the C-extension and the Python implementation are at parity.

@agronholm Now that I'm thinking about it, there might be a hacky way to make sure the C-extension and the Python implementation are identical. It would involve overloading the C-extension class to contain the dictionary mapper. Would that be acceptable? Or do you only want a modification to the C-extension?

agronholm:
Given the lack of responses, I conclude that there is no more interest in this. Closing.

PR #225, opened 2024-03-14
Moved global decoder dicts to decoder object

theeldermillenial:
This PR moves a `dict` of decoders from a global context to a class context.

The purpose of this is to make subclassing `cbor2.CBORDecoder` for modification.

This PR assists in resolving #224 as an intermediate step.

### DIFF
--- a/cbor2/_decoder.py
+++ b/cbor2/_decoder.py
@@ -217,7 +217,7 @@ class CBORDecoder:
             initial_byte = self.read(1)[0]
             major_type = initial_byte >> 5
             subtype = initial_byte & 31
-            decoder = major_decoders[major_type]
+            decoder = self.major_decoders[major_type]
             return decoder(self, subtype)
         finally:
             if immutable:
@@ -456,7 +456,7 @@ class CBORDecoder:
     def decode_semantic(self, subtype: int) -> Any:
         # Major tag 6
         tagnum = self._decode_length(subtype)
-        if semantic_decoder := semantic_decoders.get(tagnum):
+        if semantic_decoder := self.semantic_decoders.get(tagnum):
             return semantic_decoder(self)
 
         tag = CBORTag(tagnum, None)
@@ -475,7 +475,7 @@ class CBORDecoder:
 
         # Major tag 7
         try:
-            return special_decoders[subtype](self)
+            return self.special_decoders[subtype](self)
         except KeyError as e:
             raise CBORDecodeValueError(
                 "Undefined Reserved major type 7 subtype 0x%x" % subtype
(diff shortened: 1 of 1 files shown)


theeldermillenial:
@agronholm could you please take a look at this?

agronholm:
This PR only seems to do this for the pure Python variant, not the C decoder?

agronholm:
Well, that's not good. I can't have the two different implementations deviate any further.

theeldermillenial:
Does this implementation do anything that would result in a meaningful difference between Python and C?

As far as I can tell, it does not. I'm not changing anything about how the decoder actually works. I'm just changing the scope of the function calling dictionary. There are obvious differences between the Python and C versions strictly because they are written in different languages, right? I don't see how changing the scope of the function calling dict would be any different.

theeldermillenial:
If you want to suggest things I can do to align the C version of the code with Python, I am happy to implement it.

agronholm:
I think the proper place to start might be adding these new properties to `CBORDecoder_getsetters` in `decoder.c`. I don't have much bandwidth to help you with general C extension development, however. The C code was mostly contributed by another developer anyway.

theeldermillenial:
I have developed c-extensions in the past, so that's not an issue. I can contribute for sure, but I don't think a contribution make sense here because this seems to be just a language difference.

I just don't see the functional difference for the scope change because the function dictionary is only ever internally called. The current way that the Python code was written, having that dictionary be external created issues if you wanted to subclass it (which you can't do with the C-extension anyway).

If you're really looking for feature parity between these two things, the way to approach it would be a substantial rewrite of either the Python code or the C-extension.
1. On the Python side, you would want to eliminate all classes and make it a functional library.
2. Alternatively, you could turn the C-extension into a c++ extension and then you could use pybind (which is what I would recommend if you really want true feature parity).

Regardless, the scoping of that dictionary is arbitrary given the current implementation of both the C and Python sections of the code.

Please correct me if I'm wrong here, but this seems like an arbitrary change.

agronholm:
> The current way that the Python code was written, having that dictionary be external created issues if you wanted to subclass it (which you can't do with the C-extension anyway).

Wait, what?
```pyconsole
>>> import _cbor2
>>> from _cbor2 import CBORDecoder
>>> class A(CBORDecoder):
...   pass
... 
>>> dir(A)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'decode', 'decode_array', 'decode_bigfloat', 'decode_bytestring', 'decode_datetime_string', 'decode_epoch_datetime', 'decode_float16', 'decode_float32', 'decode_float64', 'decode_fraction', 'decode_from_bytes', 'decode_ipaddress', 'decode_ipnetwork', 'decode_map', 'decode_mime', 'decode_negative_bignum', 'decode_negint', 'decode_positive_bignum', 'decode_rational', 'decode_regexp', 'decode_self_describe_cbor', 'decode_semantic', 'decode_set', 'decode_shareable', 'decode_sharedref', 'decode_simple_value', 'decode_special', 'decode_string', 'decode_stringref', 'decode_stringref_namespace', 'decode_uint', 'decode_uuid', 'fp', 'immutable', 'object_hook', 'read', 'set_shareable', 'str_errors', 'tag_hook']
```
If you subclass the C `CBORDecoder` class, and it has setters/getters for these decoder maps, then isn't that what you want?

theeldermillenial:
My apologies. I'll take a look at this.

agronholm:
It seems like there's a deeper problem that allows the failure to build the C extension to pass unnoticed. I'm making the appropriate changes to the tox and GitHub actions workflows to prevent this from happening in the future.

agronholm:
Problem solved.

B3nac:
What exactly needs to be ported to `decoder.c` @agronholm?  

> I think the proper place to start might be adding these new properties to `CBORDecoder_getsetters` in `decoder.c`.

If I have an idea what needs to be added to `CBORDecoder_getsetters` I can give it a try. :rocket: 

The main reason I'm confused is the use of `self` so that the Python object can reference the dictionaries. I'm guessing all the getters would have to be moved to `CBORDecoder_getsetters` to match this and that way the two implementations don't deviate? Because that way I think in the C implimentation it wouldn't need the class prefix like `CBORDecoder.` for the getters.

agronholm:
> What exactly needs to be ported to `decoder.c` @agronholm?
> 
> > I think the proper place to start might be adding these new properties to `CBORDecoder_getsetters` in `decoder.c`.
> 
> If I have an idea what needs to be added to `CBORDecoder_getsetters` I can give it a try. 🚀
> 
> The main reason I'm confused is the use of `self` so that the Python object can reference the dictionaries. I'm guessing all the getters would have to be moved to `CBORDecoder_getsetters` to match this and that way the two implementations don't deviate? Because that way I think in the C implimentation it wouldn't need the class prefix like `CBORDecoder.` for the getters.

Uhh...I haven't touched the code base in a while (and I didn't write the C extension myself) so I'm not sure, sorry! I can take a look and get back to you, but it might take me a while.

PR #227, opened 2024-03-21
Expose `*_encoders` and use them to replace `canonical` param of `CBOREncoder`

### LINKED ISSUE #226: Re-expose `default_encoders` and allow changing them in `CBOREncoder` again
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

Please, re-expose `default_encoders` and probably other similar attributes that were public before version [5.5.0](https://github.com/agronholm/cbor2/releases/tag/5.5.0) and allow overriding `default_encoders` in  `CBOREncoder`. A new `__init__` parameter for it would be nice.

### Use case

Before [5.5.0](https://github.com/agronholm/cbor2/releases/tag/5.5.0) in [pwebarc](https://github.com/Own-Data-Privateer/pwebarc) I did

```
encoders = cbor2.encoders.default_encoders.copy()
if abridged:
    encoders[bytes] = _t.cast(_t.Any, self.encode_cbor_abridged)
    encoders[str] = _t.cast(_t.Any, self.encode_cbor_abridged)
self.encoder = _cbor2.CBOREncoder(_io.BytesIO(), default = self.encode_cbor)
self.encoder._encoders = encoders
```

(I want to use `self.encode_cbor_abridged` for long strings and bytes when abbreviation is enabled, `default_encoders` for all the default things, and `self.encode_cbor` for my custom stuff)

after 5.5.0 `default_encoders` and such are no longer exposed as a public interface and `CBOREncoder()._encoders` is now read-only.

oxij:
This changes the API a little, but makes many things much simpler. Fixes #226.

I also did some more changes to make the C module compile and all the bundled unit tests pass, see [there](https://github.com/agronholm/cbor2/compare/master...oxij:cbor2:tests-work-practice-fails?expand=1). But fixing the C module actually makes things worse in practice, because the Python module has some other nice API changes that have to be either re-implemented in C now (see my attempt at implementing `CBOREncoder_encode_container` there, but there's more API missing) or, alternatively, the design of the C module needs to change.
Personally I would prefer the second option: IMHO, it should only implement the low-level encoders and leave the rest to the Python code.

Feel free to take and edit these changes however you like, I'd agree to any other alternative solution to #226, because otherwise I'd have to vendor `cbor2` in [pwebarc](https://github.com/Own-Data-Privateer/pwebarc), which would be annoying.


agronholm:
There's a similar PR (#225) ongoing for decoders, so I'm willing to consider this. I have precious little bandwidth for this project though, so the test failures should be fixed at least.

theeldermillenial:
Hey, I authored #227 

I like your idea. I don't see a need to have a complete feature parity C-extension. Ideally the C-extension only implements the things that benefit compiled code, but obviously this is up to @agronholm 

Maybe to make things more consistent and save the repo maintainer some headaches, we could merge efforts between my PR and yours? They are functionally trying to do the same thing, just for encoders vs decoders.

agronholm:
> Maybe to make things more consistent and save the repo maintainer some headaches, we could merge efforts between my PR and yours? They are functionally trying to do the same thing, just for encoders vs decoders.

Totally. The C extension was added during a time I had delegated maintainership authority to another person. I'm all for narrowing down the C extension strictly to what benefits performance.

agronholm:
The `rust` branch redesigns the API, making this PR irrelevant. Let me know if you'd like to pursue API changes relevant to that.

PR #247, opened 2024-10-10
Drop support for EOL Python 3.8

hugovk:
## Changes

Drop support for EOL Python 3.8.

Remove redundant code and upgrade syntax for 3.9+.

https://discuss.python.org/t/python-3-8-is-now-officially-eol/66983

## Checklist

- [n/a] You've added tests (in `tests/`) added which would fail without your patch
- [x] You've updated the documentation (in `docs/`, in case of behavior changes or new
features)

these instructions.

### Updating the changelog

If, say, your patch fixes issue #999, the entry should look like this:

`* Fix big bad boo-boo in the encoder (#999
<https://github.com/agronholm/cbor2/issues/999>_; PR by @yourgithubaccount)`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,7 +26,7 @@ jobs:
       uses: pypa/cibuildwheel@v2.21.3
       env:
         CBOR2_BUILD_C_EXTENSION: "1"
-        CIBW_SKIP: pp* cp36* cp37*
+        CIBW_SKIP: pp*
         CIBW_ARCHS: auto64
         CIBW_ARCHS_MACOS: x86_64 arm64
         CIBW_ARCHS_LINUX: x86_64 aarch64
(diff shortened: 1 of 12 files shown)


.github/workflows/publish.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       uses: pypa/cibuildwheel@v2.21.3
       env:
         CBOR2_BUILD_C_EXTENSION: "1"
-        CIBW_SKIP: pp* cp36* cp37*
+        CIBW_SKIP: pp*
```
hugovk, line 29: cibuildwheel looks at `requires-python` in `pyproject.toml` so we don't need to explicitly list the old ones here.

https://cibuildwheel.pypa.io/en/stable/options/#requires-python

agronholm: Perhaps that has changed in a while because I added these specifically to block the creation of wheels for these versions after testing first. Or I may misremember and maybe I just added these based on a potential misunderstanding of their docs.

hugovk: I have some vague recollection of that too.

This feature was added in https://github.com/pypa/cibuildwheel/pull/536 which was released in https://github.com/pypa/cibuildwheel/releases/tag/2.5.0 in 2022, so not that long ago.

agronholm:
Thanks.

PR #256, opened 2025-03-26
Support encoding indefinite containers

CZDanol:
## Changes
- Added the `indefinite_containers` parameter to the encoder functions. If the parameter is set to True, containers (maps and arrays) are encoded an indefinite containers.

I am using this library to for unittesting a C++ CBOR generator, for which case I needed to generate structures with indefinite containers to be able to do 1:1 byte checks.

### DIFF
--- a/source/encoder.c
+++ b/source/encoder.c
@@ -152,7 +152,7 @@ CBOREncoder_init(CBOREncoderObject *self, PyObject *args, PyObject *kwargs)
         self->string_namespacing = true;
     }
     if (indefinite_containers == 1)
-	    self->indefinite_containers = true;
+        self->indefinite_containers = true;
 
 
     if (_CBOREncoder_set_fp(self, fp, NULL) == -1)
--- a/source/encoder.c
+++ b/source/encoder.c
@@ -113,6 +113,7 @@ CBOREncoder_new(PyTypeObject *type, PyObject *args, PyObject *kwargs)
         self->shared_handler = NULL;
         self->string_referencing = false;
         self->string_namespacing = false;
+        self->indefinite_containers = false;
     }
     return (PyObject *) self;
 }
@@ -126,16 +127,16 @@ CBOREncoder_init(CBOREncoderObject *self, PyObject *args, PyObject *kwargs)
 {
     static char *keywords[] = {
         "fp", "datetime_as_timestamp", "timezone", "value_sharing", "default",
-        "canonical", "date_as_datetime", "string_referencing", NULL
+        "canonical", "date_as_datetime", "string_referencing", "indefinite_containers", NULL
     };
     PyObject *tmp, *fp = NULL, *default_handler = NULL, *tz = NULL;
     int value_sharing = 0, timestamp_format = 0, enc_style = 0,
-	date_as_datetime = 0, string_referencing = 0;
+	date_as_datetime = 0, string_referencing = 0, indefinite_containers = 0;
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|pOpOppp", keywords,
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|pOpOpppp", keywords,
                 &fp, &timestamp_format, &tz, &value_sharing,
                 &default_handler, &enc_style, &date_as_datetime,
-                &string_referencing))
+                &string_referencing, &indefinite_containers))
         return -1;
     // Predicate values are returned as ints, but need to be stored as bool or ubyte
     if (timestamp_format == 1)
@@ -150,6 +151,8 @@ CBOREncoder_init(CBOREncoderObject *self, PyObject *args, PyObject *kwargs)
         self->string_referencing = true;
         self->string_namespacing = true;
     }
+    if (indefinite_containers == 1)
+	    self->indefinite_containers = true;
 
 
     if (_CBOREncoder_set_fp(self, fp, NULL) == -1)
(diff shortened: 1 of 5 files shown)


agronholm:
How are the tests passing for you locally?

agronholm:
> I'm getting all sorts of `SKIPPED [1] tests/test_types.py:97: requires CPython` and I don't know how to resolve it :( So I am only managing to test the python implementation locally.

This message indicates that the C extension is unavailable. So likely it didn't compile successfully, or you didn't do it in the first place. How are you running the tests? `tox` is a good way to test against multiple Python versions and its configuration on `cbor2` ensures that the C extensions get built.

CZDanol:
I was just running `pytest` on python3.12.
I now tried tox, it gives me "OK" even when I have garbage in the .c files.

CZDanol:
WOOHOO! Passes the tests now :)

agronholm:
That's odd. Try setting `CBOR2_BUILD_C_EXTENSION=1` in your environment and then run `pip install -e .`. That should make sure that the compilation happens.

docs/usage.rst
```diff
@@ -114,6 +114,11 @@ a :class:`CBORTag` object like so::
 
 This will be ignored on decode and the original data content will be returned.
 
+Indefinite containers
+-----------------
+When ``indefinite_containers=True`` is passed to :func:`dump`/:func:`dumps`, the encoder will encode containers (maps and arrays) as indefinite - the length of the container is not explicitly stored in the CBOR, but a break sign is inserted instead.
+
+
```
agronholm, line 121: I'm wondering if this section is warranted; we don't have a special section for every single option.

CZDanol: Sure, removed. I was just trying to follow the checklist.

agronholm: It's okay – documenting the parameter in the docstrings is enough in this case.

agronholm:
From a cursory look, only minor code style issues remain. In Python code, I prefer a blank line after a control block if the code continues on the same level. In C code, I prefer a space between `if` and the opening parentheses. I'll give this a proper review some time later.

CZDanol:
No problem with doing these changes, but wouldn't these styling issues be rather resolved by a pre-commit formatter hook?

agronholm:
If you can point me to a Ruff rule (or a rule in another linter) that can enforce these, I would be most grateful!

CZDanol:
Don't know about the Python part, but I am quite positive there is an option for  the ifs in ClangFormat. There seems to be no autoformatting for the C code set up at the moment.

agronholm:
Something occurred to me: couldn't we simply wrap indefinite containers to mark them to be encoded as such?

CZDanol:
You mean that you would need to use a specific class other than the `dict`? I suppose that would also work, but it wouldn't be that practical for my use case - I need to encode all containers as indefinite, implicitly. I could imagine both approaches being accepted/implemented - the wrapping approach would offer granular control over how a specific element gets encoded, whereas my PR provides a general configuration for the encoder, which I believe is also valid and useful.

agronholm:
Alright, I'll take that under advisement. Reviewing this is still on my to-do list, but I don't expect any trouble from that.

source/encoder.c
```diff
@@ -345,17 +348,19 @@ CBOREncoder_write(CBOREncoderObject *self, PyObject *data)
     Py_RETURN_NONE;
 }
 
-
 static int
-encode_length(CBOREncoderObject *self, const uint8_t major_tag,
-              const uint64_t length)
+encode_length_possibly_indefinite(CBOREncoderObject *self, const uint8_t major_tag,
```
agronholm, line 352: Can you explain why you had to split this function into two parts?

CZDanol: 1. I wanted to reduce diff by keeping rest of the encode_length calls the same (not having to add the extra parameter everywhere)
2. I felt that the bool parameter would be somewhat cryptic if there was not some addition to the function name, so I created `encode_length_possibly_indefinite`. Most of the use cases don't care about indefinite encoding, so it didn't make sense to drag this function there - so we have two now.

agronholm: I took a better look at this, and it doesn't make a whole lot of sense. Since `uint64_t` cannot possibly be -1 since it's unsigned. The actual value would then become a very large integer (18446744073709551615) instead, yes? While I doubt anyone will really try to actually encode such enormous structures, it does raise other questions, like if you felt this was an acceptable sentinel value, why then did you have to split `encode_length()` instead of just using -1 as the sentinel for indefinite length? And why can't `encode_length()` just directly look at `self->indefinite_containers` and if it's true, ignore the length parameter?

CZDanol: encode_length cannot directly use indefinite_containers, because it is also used on other places outside of encoding array/map length.

source/encoder.c
```diff
@@ -377,21 +382,64 @@ encode_length(CBOREncoderObject *self, const uint8_t major_tag,
...
+    }
+}
 
 // CBOREncoder.encode_length(self, major_tag, length)
 static PyObject *
 CBOREncoder_encode_length(CBOREncoderObject *self, PyObject *args)
 {
     uint8_t major_tag;
-    uint64_t length;
+    uint64_t length = -1;
```
CZDanol, line 413: Yes, -1 becomes a very large integer. 

I did this because I was too lazy to set up a custom structure for the `uint64_or_none` that would be required for `PyArg_ParseTuple`. It would be cleaner to have it distinct, yes. Should I remake it?

CZDanol:
Hello, no updates from you for quite some time

agronholm:
Ah...sorry, I've taken a look at this a few times but my attention span wasn't quite long enough to determine what I want here. I'll reply in the ongoing convo.

source/encoder.c
agronholm: Well, I just thought that if using -1 was acceptable, why not just use it directly in `encode_length()`?

CZDanol: I guess I can - if you accept that solution, I can make it.

agronholm: Do you see any downsides when compared to the current code in this PR?

CZDanol: At the current state of the code, probably not.

CZDanol: Updated.

CZDanol:
Okay, so apparently in the tests you are trying to encode the maximum length (u64-1) on a few places, which is now producing invalid results - so we need that discernation. I will revert the changes (and make the uint64_t_or_none proper)

agronholm:
> Okay, so apparently in the tests you are trying to encode the maximum length (u64-1) on a few places, which is now producing invalid results - so we need that discernation. I will revert the changes (and make the uint64_t_or_none proper)

I think that'd be for the best. I sometimes lose sleep over details like this :)

CZDanol:
All right, done. Sorry about the mess. I have squashed things up because I felt like it's way too much.

agronholm:
I'm too attention deficient to wrangle any more about this, so I just merged it.

CZDanol:
Okay, thanks :D

PR #267, opened 2025-12-18
fix nested shareable in arrays (python decoder only)

andreer:
## Changes

The Python decoder's decode_array() was not passing unshared=True when decoding array items, unlike decode_map() which correctly does this. This allowed nested shareable tags (tag 28) inside arrays to overwrite _share_index, causing subsequent shared references (tag 29) to resolve to the wrong object.

```
# [Tag 28 -> [Tag 28 -> 'a', 'b'], Tag 29 -> 0]
data = bytes.fromhex('82d81c82d81c61616162d81d00')
loads(data)
# Before: [['a', 'b'], 'b']  - shared ref returns 'b' instead of the array
# After:  [['a', 'b'], ['a', 'b']] - both elements are the same list object
```

The C decoder already handles this correctly.

## Checklist

features

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -9,6 +9,7 @@ This library adheres to `Semantic Versioning <https://semver.org/>`_.
 
 - Fixed Python decoder not preserving share index when decoding array items containing
   nested shareable tags, causing shared references to resolve to wrong objects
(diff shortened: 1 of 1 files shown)


agronholm:
Do you have more PRs coming in the short term, or should I cut a new release now?

PR #204, opened 2023-12-30
Fixed MemoryError when decoding large definite strings

agronholm:
Relates to #198.

PR #271, opened 2026-01-03
Fix str_errors parameter regression in C extension (#255)

### LINKED ISSUE #255: `loads` `str_errors` kwarg no longer worker since 5.6.0
### Things to check first

### cbor2 version

5.6.5

### Python version

Python 3.10.14

### What happened?

The `str_errors` kwarg is no longer working as expected since 5.6.0:
When passing `str_errors="replace"`, bytes that failed to decode, should be replaced with a replacement character. Instead, `CBORDecodeValueError` is always raised.

Before 5.6.0 (e.g. 5.5.1):
```
>>> cbor2.loads(b"cfo\x90", str_errors="replace")
'fo�'
```

Since 5.6.0:
```
>>> cbor2.loads(b"cfo\x90", str_errors="replace")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x90 in position 2: invalid start byte

The above exception was the direct cause of the following exception:

_cbor2.CBORDecodeValueError: error decoding unicode string
```

topher200:
## Summary

Fixes #255 - Restores support for the `str_errors` parameter in the C extension when decoding CBOR strings containing invalid UTF-8 byte sequences. The parameter has been broken since v5.6.0.

## Problem Description

Since version 5.6.0, the `str_errors` parameter in `cbor2.loads()` stopped working in the C extension (`_cbor2`). When decoding CBOR data containing invalid UTF-8 sequences:

**Expected behavior (v5.5.1 and earlier):**
```python
cbor2.loads(b"cfo\x90", str_errors="replace")  # Returns: 'fo�'
```

**Broken behavior (v5.6.0 - v5.8.0):**
```python
cbor2.loads(b"cfo\x90", str_errors="replace")  # Raises: CBORDecodeValueError
```

The pure Python implementation (`cbor2._decoder`) continued to work correctly; only the C extension was affected.

## Root Cause

**Commit 387755e (PR #204) - "Fixed MemoryError when decoding large definite strings"** Date: Jan 14, 2024

This commit rewrote the string decoding logic in `source/decoder.c` to handle large strings in chunks, fixing memory issues. However, it accidentally removed the `str_errors` parameter from UTF-8 decoding calls:

**Before (v5.5.1 - working):**
```c
ret = PyUnicode_DecodeUTF8(buf, length, PyBytes_AS_STRING(self->str_errors));
```

**After (v5.6.0 - broken):**
```c
// Short strings (≤65536 bytes)
PyObject *ret = PyUnicode_FromStringAndSize(bytes, length);  // No error handler support

// Long strings (>65536 bytes)
string = PyUnicode_DecodeUTF8Stateful(source_buffer, chunk_length, NULL, &consumed);  // NULL instead of str_errors
```

### Changes

Re-add the missing parameter.

Fixes #255 

 ### Test Plan

Add a new unit test for this functionality.

## Checklist

features

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,12 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed ``str_errors`` parameter being ignored in C extension when decoding strings with
+  invalid UTF-8 byte sequences (broken since 5.6.0)
+  (`#255 <https://github.com/agronholm/cbor2/issues/255>`_; PR by @topher200)
+
 **5.8.0** (2025-12-30)
 
 - Added readahead buffering to C decoder for improved performance.
--- a/source/decoder.c
+++ b/source/decoder.c
@@ -836,7 +836,7 @@ decode_definite_short_string(CBORDecoderObject *self, Py_ssize_t length)
         return NULL;
 
     const char *bytes = PyBytes_AS_STRING(bytes_obj);
-    PyObject *ret = PyUnicode_FromStringAndSize(bytes, length);
+    PyObject *ret = PyUnicode_DecodeUTF8(bytes, length, PyBytes_AS_STRING(self->str_errors));
     Py_DECREF(bytes_obj);
     if (ret && string_namespace_add(self, ret, length) == -1) {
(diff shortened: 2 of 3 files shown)


agronholm:
@topher200 Perhaps you should subscribe to releases of this project? Github lets you do that.

PR #275, opened 2026-01-19
Set default read_size to 1 for backwards compatibility

### LINKED ISSUE #272: Sequential decoding does not behave as expected
### Things to check first

### cbor2 version

5.8.0

### Python version

3.12.3

### What happened?

In cbor2 before version 5.8 I could construct a `BytesIO` buffer and sequentially decode items using a `CBORDecoder` object. Starting in 5.8 it appears that the decoder `fp` member is seeked to the end of the buffer and calling `dec.fp.read()` gives empty bytes because of that. In earlier versions, the decoder would leave the underlying `fp` cursor just past the last decoded CBOR item.

Here is a small script that succeeds with earlier versions and fails on line 9 with 5.8.0, where the cursor is reported as position 3 (EOF).

```py
import cbor2
import io

data = bytes.fromhex('000102')

dec = cbor2.CBORDecoder(io.BytesIO(data))
assert 0 == dec.fp.tell()
assert 0 == dec.decode()
assert 1 == dec.fp.tell()
assert 1 == dec.decode()
assert 2 == dec.fp.tell()
assert bytes.fromhex('02') == dec.fp.read()
```

andreer:
## Changes

Set default read_size to 1.

Fixes #272

The buffered reads introduced in 5.8.0 could cause issues when code needs to access the stream position after decoding. This changes the default back to 1 (matching 5.7.1 behavior) while still allowing users to opt-in to faster decoding by passing read_size=4096.

Implementation details:
- Use function pointer dispatch to eliminate runtime checks for read_size=1
- Skip buffer allocation entirely for unbuffered path
- Add read_size parameter to load() and loads() for API completeness

## Checklist

- [X] You've updated the documentation (in `docs/`), in case of behavior changes or new
features

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -74,6 +74,17 @@ instead encodes a reference to the nth sufficiently long string already encoded.
 .. warning:: Support for string referencing is rare in other CBOR implementations, so think carefully
     whether you want to enable it.
 
+Performance tuning
+------------------
+
+By default, the decoder only reads the exact amount of bytes it needs. But this can negatively
+impact the performance due to the potentially large number of individual read operations.
+To make it faster, you can pass a different ``read_size`` parameter (say, 4096), to :func:`load`,
+:func:`loads` or :class:`CBORDecoder`.
+
+.. warning:: If the input stream contains data other than the CBOR stream, that data (or parts of)
+    may be lost.
+
 Tag support
 -----------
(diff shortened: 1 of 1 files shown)


agronholm:
For anyone watching, I think I'll finish the code and docs myself in a few days.

PR #278, opened 2026-02-04
Rust conversion

### LINKED ISSUE #41: The encoder is difficult to customize
Right now the only simple way to avoid #37 is to modify a global variable (see code below), which could theoretically break another module that uses `cbor2` in the same process and expects the encoding of `date` objects to work.

```python
import cbor2
cbor2.encoder.default_encoders.pop(date)
```

It could be nice to have a `CBOR` class that would hold configuration variables and implement the `dump`, `dumps`, `load` and `loads` functions. That would make it possible to create a custom CBOR encoder/decoder that could be used exactly like the global `cbor2.*` functions but would encode/decode slightly differently. For example:

```python
from datetime import date
from cbor2 import CBOR
dateless_cbor = CBOR()
dateless_cbor.encoders.pop(date, None)
dateless_cbor.dumps(date(2019, 2, 20))  # raises CBOREncodeError
```

### LINKED ISSUE #157: Bug in `pack_float16()`
I did some cursory testing of your float16 functions (out of curiosity, since I recently wrote my own conversion functions float16 <-> float32).

It seems your `pack_float16()` doesn't give the same answer as mine for some values. For small numbers, it could maybe be attributed to different rounding rules, but for high numbers it does seem wrong, for example for the float32 number represented in raw hex as `0x47000000`, i.e. `32768.0`, which has an exact representation in float16: `0x7800`, but is rounded to infinity by `pack_float16()`.

I did read the document you referenced in `half_float_tables.py`, and it does not specify any rounding mode, so I don't know how it handles that. My code is written to do "Round to nearest, ties to even", which I tested against numpy.

I looked at the code and realized that `pack_float16()` being wrong is not critical, I just wanted to let you know.

If you're curious, my code and tests can be found here: https://github.com/NordicSemiconductor/zcbor/pull/285

### LINKED ISSUE #169: gcc compilation warnings
```
x86_64-pc-linux-gnu-gcc -Wsign-compare -DNDEBUG -O2 -pipe -march=core2 -DNDEBUG -fPIC -I/usr/include/python3.11 -c source/decoder.c -o build/temp.linux-x86_64-cpython-311/source/decoder.o -std=c99 -D_GNU_SOURCE
source/decoder.c: In function ‘decode_bytestring’:
source/decoder.c:605:33: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘uint64_t’ {aka ‘long unsigned int’} [-Wformat=]
  605 |         sprintf(length_hex, "%llX", length);
      |                              ~~~^   ~~~~~~
      |                                 |   |
      |                                 |   uint64_t {aka long unsigned int}
      |                                 long long unsigned int
      |                              %lX
source/decoder.c: In function ‘decode_string’:
source/decoder.c:706:33: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘uint64_t’ {aka ‘long unsigned int’} [-Wformat=]
  706 |         sprintf(length_hex, "%llX", length);
      |                              ~~~^   ~~~~~~
      |                                 |   |
      |                                 |   uint64_t {aka long unsigned int}
      |                                 long long unsigned int
      |                              %lX
source/decoder.c: In function ‘decode_array’:
source/decoder.c:863:33: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘uint64_t’ {aka ‘long unsigned int’} [-Wformat=]
  863 |         sprintf(length_hex, "%llX", length);
      |                              ~~~^   ~~~~~~
      |                                 |   |
      |                                 |   uint64_t {aka long unsigned int}
      |                                 long long unsigned int
      |                              %lX
x86_64-pc-linux-gnu-gcc -Wsign-compare -DNDEBUG -O2 -pipe -march=core2 -DNDEBUG -fPIC -I/usr/include/python3.11 -c source/encoder.c -o build/temp.linux-x86_64-cpython-311/source/encoder.o -std=c99 -D_GNU_SOURCE
source/encoder.c: In function ‘CBOREncoder_init’:
source/encoder.c:187:36: warning: initialization of ‘PyObject *’ {aka ‘struct _object *’} from incompatible pointer type ‘PyTypeObject *’ {aka
‘struct _typeobject *’} [-Wincompatible-pointer-types]
  187 |         PyObject *datetime_class = PyDateTimeAPI->DateType;
      |                                    ^~~~~~~~~~~~~

```
Building from Gentoo ebuild https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-python/cbor2/cbor2-5.4.6.ebuild
gcc 12.2.1

### LINKED ISSUE #192: C-extension does allow customizing serialization
### Things to check first

### cbor2 version

5.5.1

### Python version

3.11

### What happened?

The behaviour of the serializer differs between the C extension and the Python implementation of this library.

The Python implementation allows customizing the by updating the value of the `_encoders` map for known types. The C-implementation as an optimization does not use this map and defines it's own implementations of encoders for the known types. [This comment](https://github.com/agronholm/cbor2/issues/47#issuecomment-491543272) on the original issue adding the C implementation shows that it was intended that the C implementation allow for the `canonical` parameter to be set to a value of 2 to opt into the slower codepath where the Python defined encoders were used.

The two changes below appear to be the cause of this:

* https://github.com/agronholm/cbor2/pull/77 enforced that `enc_style` is only ever 0 or 1 (0 when initialized, 1 if the input is exactly 1).
* [This commit](https://github.com/Sekenre/cbor2/commit/9927bb580950b61845158f5a1f0658aa08e4edec) changed the parsing of the field to parse it as a predicate and not a byte.

A few questions here:
1. I recognize that modifying `_encoders` is not necessarily a supported interface for the library given how the variable is named? If that's not the case, is there no supported way to perform this substitution?
2. If allowing the C implementation to use the Python defined encoders, how should this be exposed? Go back to overloading the `canonical` parameter? Introduce an Enum as a value for that field?

The following test case should show the difference.
The python implementation will pass, the C implementation will fail.

```
import calendar

def test_encoder_replacement(impl):
    def to_epoch(dt: datetime) -> float:
        millis = int(dt.microsecond / 1000.0) / 1000.0
        return calendar.timegm(dt.astimezone(timezone.utc).timetuple()) + millis


    def encode_datetime(self, obj: datetime):
        self.encode_float(to_epoch(obj))


    dt = datetime(2013, 3, 21, 20, 4, 0, tzinfo=timezone.utc)
    expected = unhexlify("fb41d452d9ec000000")
    with BytesIO() as fp:
        encoder = impl.CBOREncoder(fp, canonical=2)
        encoder._encoders[datetime] = encode_datetime
        encoder.encode(dt)
        serialized = fp.getvalue()
 ```

### LINKED ISSUE #224: Better support for lists

### LINKED ISSUE #226: Re-expose `default_encoders` and allow changing them in `CBOREncoder` again

### LINKED ISSUE #232: Support of RFC9164 Semantic Tags for IP Addresses/Prefixes
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

[RFC9164](https://www.rfc-editor.org/rfc/rfc9164.html) changes the semantic tags for IP Addresses and prefixes. See the RFC for specific details.

TLDR: Instead of 260 and 261 there is now semantic tags 52 (for IPv4) and 54 (for IPv6). Each tag supports a combination of full address and prefix forms. The originals are [deprecated](https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml).

### Use case

[IANA](https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml) has, per RFC9164, marked the original semantic tags as deprecated in favor of the RFC9164 tags.

While I am unsure of how much these new tags are used in the wild, I am currently proposing their use in various documents for the Drone Remote ID Protocol (DRIP).  For some examples see [this document](https://www.ietf.org/archive/id/draft-ietf-drip-registries-15.html), Section 9 and Appendix A.

### LINKED ISSUE #257: Memory leak in C extension when decoding fails inside nested map or string
### Things to check first

### cbor2 version

5.6.6.dev1

### Python version

3.12

### What happened?

When failing to decode certain corrupted types of data there is a memory leak in the C extension. When running the pure python implementation with `CBOR2_BUILD_C_EXTENSION=0` this **does not** happen. The corrupted type I was able to reproduce was failing to decode a string inside of a nested map. It fails here in the pure python implementation for reference. 

```
            if length <= 65536:
                try:
                    result = self.read(length).decode("utf-8", self._str_errors)
                except UnicodeDecodeError as exc:
                    raise CBORDecodeValueError("error decoding unicode string") from exc
```
This `CBORDecodeValueError` path in the C extension would likely be the culprit where it exits without cleanup. Possibly a missing `Py_DECREF` in the failure path.

cbor2_repro.py
```python
import cbor2

# malformed string at the leaf of nested maps
bad_cbor = (
    b'\xbf' +            # indefinite map
    b'\x01' +            # key = 1
    b'\xa1' +            # value = map of 1 pair
    b'\x02' +            # key = 2
    b'\x63foo'[:-1] +    # string "foo", but truncated → invalid UTF-8
    b'\xff'              # break
)

for _ in range(10_000_000):
    try:
        cbor2.loads(bad_cbor)
    except Exception:
        pass
```
Memory usage will steadily grow under mprof. Below are memory usage for the pure python implementation showing no memory usage increase over time, but continuous increase in usage under the c extensions version.

Image
Image

`mprof run python cbor2_repro.py`

### LINKED ISSUE #272: Sequential decoding does not behave as expected

### LINKED ISSUE #285: CBORDecoder.read(n) raises MemoryError on large n instead of CBORDecodeEOF
### Things to check first

### cbor2 version

5.8.0

### Python version

3.9.18

### What happened?

When `CBORDecoder.read(n)` is called with `n` exceeding the actual stream length, a slightly large `n` (e.g. `10`) correctly raises `_cbor2.CBORDecodeEOF`. However, a large `n` (e.g. `10**18`) does not reach this check, it raises `MemoryError` instead.

```python
import io
import cbor2

dec = cbor2.CBORDecoder(fp=io.BytesIO(b"111"), str_errors="error")

# 抛出 CBORDecodeEOF
dec.read(10)

# 抛出 MemoryError
dec2 = cbor2.CBORDecoder(fp=io.BytesIO(b"111"), str_errors="error")
dec2.read(10**18)
```

Traceback:

`n = 10` 时：

```
  File "test.py", line 38, in <module>
    ret = obj.read(10)
_cbor2.CBORDecodeEOF: premature end of stream (expected to read 10 bytes, got 3 instead)
```

`n = 10**18` 时：

```
  File "test.py", line 38, in <module>
    ret = obj.read(1000000000000000000)
MemoryError
```

Expected behavior :`CBORDecoder.read(10**18)` should raise `_cbor2.CBORDecodeEOF`(same as `read(10)`), not `MemoryError`.

### LINKED ISSUE #286: Decoding CBOR tag 5  with `null` mantissa raises TypeError instead of CBORDecodeValueError
### Things to check first

### cbor2 version

5.8.0

### Python version

3.9.18

### What happened?

When decoding CBOR tag 5 (bigfloat) with a null mantissa, `CBORDecoder.decode()` raises` TypeError` instead of the expected `CBORDecodeValueError.` This is probably because the try/except in decode_bigfloat only covers `self._decode() `, leaving the `Decimal(sig) * (2 ** Decimal(exp)) `computation outside its scope.

```
data = bytes.fromhex("c58200f6")
obj = cbor2.CBORDecoder(io.BytesIO(data))
print(obj.decode())
```

trace:
```
  File "test.py", line 60, in <module>
    print(obj.decode())
TypeError: unsupported operand type(s) for *: 'NoneType' and 'decimal.Decimal'

```

Expected behavior: should raise CBORDecodeValueError, consistent with other invalid input.

agronholm:
## Changes

This replaces the encoder and decoder code (Python and C) with a new Rust-based implementation. This was done to
reduce the maintenance cost (by having only a single implementation) and to prevent memory corruption
issues and vulnerabilities that have been appearing in the C version.

The Rust version also supports free-threading and subinterpreters out of the box thanks to native multi-stage initialization
and synchronization.

There is also a significant performance boost in both decoding and encoding, up to 400%, partly due to
using internal buffers more than reading/writing from/to files which cause round-trips to the Python interpreter.

Finally, it implements tags 52 and 54 for the official IP address/network/interface
specification. The deprecated tags 260 and 261 are still supported but only for decoding.
Additionally, tag 261 will now decode to a `IPv4Interface` or `IPv6Interface` if there are host bits present in the address.

Fixes #286. Fixes #285. Fixes #272. Fixes #257. Closes #232. Closes #226. Closes #224. Closes #192. Closes #169. Fixes #157. Closes #41.

## Checklist

- [X] You've updated the documentation (in `docs/`), in case of behavior changes or new
features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/test_decoder.py
+++ b/tests/test_decoder.py
@@ -45,12 +45,11 @@ from cbor2 import (
     shareable_decoder,
     undefined,
 )
-from cbor2._cbor2 import loads as raw_loads
 
 if sys.hexversion < 51314855:
     from cbor2 import frozendict
 
-DECODER_MAX_DEPTH = signature(raw_loads).parameters["max_depth"].default
+DECODER_MAX_DEPTH = signature(loads).parameters["max_depth"].default
 
 
 @pytest.fixture


agronholm:
My analysis shows that I'm handling the read buffers inefficiently and copying around too much memory.

PR #298, opened 2026-05-02
Broadened decoder support to any type supporting the buffer interface

### LINKED ISSUE #297: New Rust implementation doesn't accept memoryview
### Things to check first

### cbor2 version

6.0.1

### Python version

3.13

### What happened?

The new Rust implementation doesn't accept bytes-like objects. I received this exception via the webauthn package:

```
>           to_return = cbor2.loads(data)
                        ^^^^^^^^^^^^^^^^^
E           TypeError: argument 'data': 'memoryview' object is not an instance of 'bytes'
```

I don't really know much about language bindings - but cbor2 is a python library and as such it would be great if it accepted 'duck-typed' input (which the older implementation did - and in fact explicitly accepted [bytes, bytearray, memoryview]).

Possibly cbor2.loads() should be a python shim that casts/converts python bytes-like input to Rusts strict bytes input.

```
import cbor2
v = memoryview(b'mycbordata')
cbor2.loads(v)

  File "<input>", line 1, in <module>
TypeError: argument 'data': 'memoryview' object is not an instance of 'bytes'
```

agronholm:
## Changes

Broadens input type support to all objects implementing the buffer protocol.

Closes #297.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -10,7 +10,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Added the ``allow_duplicate_keys`` parameter to :class:`CBORDecoder`, :func:`load` and
   :func:`loads` (default: ``True``). When set to ``False``, a :exc:`CBORDecodeError` is raised
   upon encountering a duplicate key within the same map.
-  (`#283 <https://github.com/agronholm/cbor2/pull/283>`_)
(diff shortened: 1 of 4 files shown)


jwag956:
Thanks - at least for the use case from webauthn - copying is no big deal - we're talking 64 bytes or so.

PR #307, opened 2026-06-02
Fixed read_exact() truncating buffer incorrectly

Noderyos:
## Changes

New code added in 985d54c introduced a bug.

Currently, in `read_exact`, when `self.available_bytes < N`, `needed_bytes` is set at `N - self.available_bytes`.
`concatenated_buffer` is then truncated to `needed_bytes` and extended with `new_bytes` of length `needed_bytes`.

However, to match the final expected length of `N`, `concatenated_buffer` should be truncated to `self.available_bytes`.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,12 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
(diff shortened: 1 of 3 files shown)


agronholm:
Thanks!

PR #308, opened 2026-06-02
fix incorrect length tracking for chunked text strings

sahvx655-wq:
decode_string reuses the `length` variable as the chunk loop counter and decrements it to 0 before returning it, so any definite-length text string over 65536 bytes reports a length of 0. That length feeds the string-reference namespace eligibility check (tag 256), so a long string is never registered, misaligning later string-reference indices and decoding valid CBOR to the wrong string or raising "string reference not found". The bytestring path already tracks the remaining length separately; mirror that here.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,12 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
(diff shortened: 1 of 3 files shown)


agronholm:
This requires a regression test and a changelog entry. Both were mentioned on the PR checklist which you deleted.

agronholm:
Thanks. But next time please respect the PR template.

PR #309, opened 2026-06-03
reject over-length address byte strings in ip network decoders

sahvx655-wq:
Spotted this reading the tag 52/54 decoders: the network branches call resize(4, 0) / resize(16, 0) on the address byte string, which is meant to pad an abbreviated address with trailing zeros. It also silently truncates a byte string that's longer than the address, so a 6-byte IPv4 or 18-byte IPv6 address decodes to a wrong network instead of being rejected. RFC 9164 says an over-length byte string here is malformed, so this now raises before the resize. Padding of shorter strings is unchanged. Test and changelog included.


agronholm (approved):
Looks good, thanks.

PR #311, opened 2026-06-05
Add array_hook to surface indefinite-length arrays on decode

theeldermillenial:
# PR 1 — Add `array_hook` to surface indefinite-length arrays on decode

**Branch:** `feature/array-hook`

## Problem
cbor2 6.x decodes definite- and indefinite-length arrays into the same Python type (`list`, or `tuple` under `immutable=True`), discarding the length-encoding distinction. There is no decoder hook or return-type difference that lets a caller observe whether a given array used indefinite-length (`0x9f … 0xff`) encoding.

This makes **byte-exact round-tripping impossible** for callers that must reproduce the original encoding — e.g. anything that decodes CBOR, transforms it, and re-encodes it for hashing or signature verification (the Cardano ecosystem relies on this). The encoder already supports emitting indefinite containers; only the decode direction lacks a way to preserve the information.

## Solution
Add an `array_hook` decoder callback, mirroring the existing `object_hook`. It is invoked for every decoded array with `(value, indefinite)`:
- `value` — the decoded array (`list`, or `tuple` under `immutable=True`)
- `indefinite` — `True` if the array used indefinite-length encoding

Its return value replaces the array in the output. `indefinite` is precisely the one piece of metadata a caller cannot otherwise recover (immutability is already inferable from the value's type), which is why it's passed explicitly.

```python
from cbor2 import loads
def array_hook(value, indefinite):
    return MyIndefiniteList(value) if indefinite else value
loads(data, array_hook=array_hook)
```

Default is `None` → today's behavior, zero breakage.

## Changes
- `rust/decoder.rs`: `array_hook` field, constructor param, getter/setter (with callable validation), and a `maybe_call_array_hook` helper applied at every `decode_array` completion site (definite → `False`, indefinite → `True`, including empty arrays).
- `rust/lib.rs`: `array_hook` threaded through `load()`/`loads()` + docstrings.
- `python/cbor2/__init__.pyi`: `ArrayHook` type alias on `CBORDecoder`, `load`, `loads`.
- `tests/test_decoder.py`: 9 tests — attribute get/set/not-callable/delete, definite, indefinite, return-replacement, nested, immutable/tuple, exception propagation.
- changelog entry.

## Verification
`cargo check --deny warnings` + clippy clean; full test suite passes (441 baseline + 9 new = 450).

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,15 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Added the ``array_hook`` parameter to :class:`CBORDecoder`, :func:`load` and :func:`loads`.
+  When set, it is called for each decoded array with two arguments: the array (a :class:`list`,
+  or a :class:`tuple` when ``immutable`` is true) and a boolean indicating whether the array used
+  indefinite-length encoding. Its return value is substituted for the array in the output. This
+  allows callers to preserve the distinction between definite- and indefinite-length arrays, which
+  is otherwise lost after decoding.
+
 **6.1.2** (2026-06-02)
 
 - Fixed incorrect tracking of string references for definite-length text strings of length greater
--- a/python/cbor2/__init__.pyi
+++ b/python/cbor2/__init__.pyi
@@ -48,6 +48,7 @@ _TContainer = TypeVar("_TContainer", default=None)
 TagHook: TypeAlias = Callable[[CBORTag, bool], Any]
 SemanticDecoderCallback: TypeAlias = Callable[[Any, bool], Any]
 ObjectHook: TypeAlias = Callable[[Mapping[Any, Any], bool], Any]
+ArrayHook: TypeAlias = Callable[[Sequence[Any], bool], Any]
 EncoderHook: TypeAlias = Callable[[CBOREncoder, Any], Any]
 ShareableDecoderCallback: TypeAlias = Callable[[Any], Any]
 ShareableDecoderInitializer: TypeAlias = Callable[[bool], tuple[Any, ShareableDecoderCallback]]
@@ -117,6 +118,7 @@ class CBORDecoder:
     fp: IO[bytes]
     tag_hook: TagHook | None
     object_hook: ObjectHook | None
+    array_hook: ArrayHook | None
     str_errors: str
     def __new__(
         cls,
@@ -124,6 +126,7 @@ class CBORDecoder:
         *,
         tag_hook: TagHook | None = ...,
         object_hook: ObjectHook | None = ...,
+        array_hook: ArrayHook | None = ...,
         semantic_decoders: Mapping[int, SemanticDecoderCallback | ShareableDecoderInitializer]
         | None = ...,
         str_errors: str = ...,
@@ -214,6 +217,7 @@ def load(
     *,
     tag_hook: TagHook | None = ...,
     object_hook: ObjectHook | None = ...,
+    array_hook: ArrayHook | None = ...,
     semantic_decoders: Mapping[int, SemanticDecoderCallback | ShareableDecoderInitializer]
     | None = ...,
     str_errors: str = ...,
@@ -228,6 +232,7 @@ def loads(
     *,
     tag_hook: TagHook | None = ...,
     object_hook: ObjectHook | None = ...,
+    array_hook: ArrayHook | None = ...,
     semantic_decoders: Mapping[int, SemanticDecoderCallback | ShareableDecoderInitializer]
     | None = ...,
     str_errors: str = ...,
--- a/rust/decoder.rs
+++ b/rust/decoder.rs
@@ -158,6 +158,13 @@ fn require_tuple<'py>(value: Bound<'py, PyAny>, length: usize) -> PyResult<Bound
 ///     callable that takes 2 arguments: the decoder instance, and a dictionary. This
 ///     callback is invoked for each deserialized :class:`dict` object. The return value
 ///     is substituted for the dict in the deserialized output.
+/// :param array_hook:
+///     callable that takes 2 arguments: the decoded array (a :class:`list`, or a
+///     :class:`tuple` when ``immutable`` is true), and a boolean that is :data:`True` if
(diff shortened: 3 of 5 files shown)


agronholm:
My long-term solution for this is to add streaming-type decoding which yields low-level items like "array start", "array end", etc. There are other use cases for this too. It is a big undertaking, though, which is why I haven't done it yet.

theeldermillenial:
That makes sense.

This particular issue is really breaking our ability to do roundtrip serialization, which is important for the reproducibility needed for hashing cbor.

Are you opposed to merging this in until you do the streaming-type decoding?

agronholm:
To be frank, this is a custom feature for your use case only. I would only consider added complexity if it was a universal feature that served a broader audience. Streaming decoding is a feature that has been requested by a number of users. So, to answer your question, I am opposed to this PR.

theeldermillenial:
I appreciate your response, and for the work you do.

I think that leaves me with no other option than to fork and maintain it.

agronholm:
Can you confirm that the streaming decoding will at least solve your problem? I was also thinking about why you're special casing arrays. Wouldn't you need to also know for text strings, byte strings and maps if they were indefinite or not when the decoder received them?

agronholm:
My idea is that the streaming decoder could function as the backing decoder for `load()` and `loads()` without any significant drop in performance as they would interface directly in Rust.

agronholm:
@theeldermillenial Okay, I have something for you now! Check out the `streaming-decoder` branch. I have a sample script to demonstrate what it can do here:


<details><summary>Demo script</summary>
<p>

```py
"""Byte-exact CBOR round-tripping via the low-level streaming decoder.

The regular high-level API (``loads``/``dumps``) throws away the
*definite vs. indefinite* length distinction: an indefinite array
``0x9f ... 0xff`` and a definite array ``0x83 ...`` both decode to a plain
``list``, so re-encoding cannot reproduce the original bytes.  For callers
that must reproduce the exact encoding -- e.g. hashing or signature
verification over CBOR (the Cardano use case in agronholm/cbor2#311) -- that
loss is fatal.

Iterating ``CBORStreamDecoder`` exposes the raw token stream, where the
distinction survives as ``ArrayStart(None)`` vs ``ArrayStart(n)``,
``TextStringStart`` vs ``TextString``, and so on.  Re-emitting those tokens
with ``CBOREncoder``'s low-level primitives reproduces the original bytes
exactly -- which was impossible before the streaming decoder existed.
"""

from __future__ import annotations

from io import BytesIO

from cbor2 import CBOREncoder, CBORStreamDecoder, CBORSimpleValue, dumps, loads
from cbor2 import tokens as t


def reencode_tokens(tokens, fp) -> None:
    """Write a token stream back out, preserving the definite/indefinite head.

    ponytail: preserves the length-encoding distinction (the thing that was
    impossible before). It does not preserve *non-minimal* integer/length
    heads or the exact float width (f16/f32/f64), since the tokens don't
    carry that; add if a producer emits non-canonical heads.
    """
    enc = CBOREncoder(fp)
    for tok in tokens:
        match tok:
            # --- structural: the bit loads()/dumps() can't preserve ---
            case t.ArrayStart(length=n):
                enc.encode_length(4, n)  # n is None for indefinite
            case t.MapStart(length=n):
                enc.encode_length(5, n)
            case t.ByteStringStart():
                enc.encode_length(2, None)
            case t.TextStringStart():
                enc.encode_length(3, None)
            case t.Tag(number=num):
                enc.encode_length(6, num)
            case t.Break():
                enc.encode_break()
            # --- leaves ---
            case t.Integer(value=v):
                enc.encode_int(v)
            case t.ByteString(value=v):
                enc.encode_bytes(v)
            case t.TextString(value=v):
                enc.encode_string(v)
            case t.Float(value=v):
                enc.encode_float(v)
            case t.Boolean(value=v):
                enc.encode_bool(v)
            case t.Simple(value=v):
                enc.encode_simple_value(CBORSimpleValue(v))
            case t.Null():
                enc.encode_none()
            case t.Undefined():
                enc.encode_undefined()
            case _:
                raise TypeError(f"unhandled token: {tok!r}")


# A hand-built document that deliberately MIXES encodings:
#   - an indefinite-length outer array (0x9f ... 0xff)
#   - a definite-length inner array [2, 3]  (0x82 ...)
#   - an indefinite-length text string "hello" split as "he" + "llo"
#   - a definite byte string
payload = bytes(
    [
        0x9F,  # array(*)  -- indefinite
        0x01,  #   1
        0x82, 0x02, 0x03,  #   [2, 3]  -- definite
        0x7F,  #   text(*) -- indefinite
        0x62, 0x68, 0x65,  #     "he"
        0x63, 0x6C, 0x6C, 0x6F,  #     "llo"
        0xFF,  #   break (end text)
        0x42, 0x01, 0x02,  #   h'0102'  -- definite byte string
        0xFF,  # break (end array)
    ]
)

value = loads(payload)
print("decoded value :", value)

# 1) High-level round-trip LOSES the encoding -> bytes differ.
naive = dumps(value)
print("original bytes:", payload.hex())
print("dumps(loads()):", naive.hex())
assert naive != payload, "expected the naive round-trip to differ"
print("naive round-trip is NOT byte-exact  (indefinite encoding lost)\n")

# 2) Streaming round-trip PRESERVES the encoding -> bytes identical.
out = BytesIO()
reencode_tokens(CBORStreamDecoder(BytesIO(payload)), out)
exact = out.getvalue()
print("token round-trip:", exact.hex())
assert exact == payload, "streaming round-trip must be byte-exact"
print("streaming round-trip IS byte-exact\n")

# 3) ...and you can still transform while preserving the encoding.
#    Here: bump every integer by 100, indefinite markers untouched.
def bump_ints(tokens):
    for tok in tokens:
        yield t.Integer(tok.value + 100) if isinstance(tok, t.Integer) else tok


out = BytesIO()
reencode_tokens(bump_ints(CBORStreamDecoder(BytesIO(payload))), out)
transformed = out.getvalue()
print("transformed    :", transformed.hex(), "->", loads(transformed))
# still indefinite array + indefinite text, only the ints changed
assert transformed[0] == 0x9F and loads(transformed) == [101, [102, 103], "hello", b"\x01\x02"]
print("transform kept the indefinite encoding")
```

</p>
</details>

agronholm:
I'm closing this PR given the more comprehensive solution I have currently in the pipeline.

Issue #295, opened 2026-04-28
Inconsistent decoding of dicts


TLCFEM:
### Things to check first

- [x] I have searched the existing issues and didn't find my bug already reported there

- [x] I have checked that my bug is still present in the latest release

### cbor2 version

6

### Python version

3.11

### What happened?

The following code snippet prints:

```text
b'\xa1at\xa1crun\xa1ap\x82\x05\x18z'
{'t': {'run': frozendict({'p': (5, 122)})}}
b'\xa1crun\xa1ap\x82\x05\x18z'
{'run': {'p': (5, 122)}}
```

```py
import cbor2


def roundtrip(origin):
    origin_encoded = cbor2.dumps(origin)
    print(origin_encoded)
    decoded = cbor2.loads(origin_encoded)
    print(decoded)


if __name__ == "__main__":
    roundtrip({"t": {"run": {"p": [5, 122]}}})
    roundtrip({"run": {"p": [5, 122]}})
```

Why is so the same dict `{"p": [5, 122]}` becomes `frozendict` in one case?

### How can we reproduce the bug?

See above

TLCFEM:
Looks like this flag is messed up.

https://github.com/agronholm/cbor2/blob/03c656c840bad666dde2562750370ac4ab6ad2f6/rust/decoder.rs#L1503

agronholm:
From the looks of it, it doesn't reset the immutable flag when it should. I'll investigate.

agronholm:
This is where it likely goes wrong: https://github.com/agronholm/cbor2/blob/master/rust/decoder.rs#L1578-L1579

agronholm:
No, that wasn't it. The problem was that the `ContinueFrame` handler didn't actually set the `immutable` flag on the frame itself, only its current (transient) value, so when the next `ContinueFrame` handler looked it up, it was a `true` value left over from key decoding of the previous frame.

Issue #296, opened 2026-04-30
Tag values always decoded as immutable


oyvindronningstad:
### Things to check first

- [x] I have searched the existing issues and didn't find my bug already reported there

- [x] I have checked that my bug is still present in the latest release

### cbor2 version

6.0.1

### Python version

3.14.4

### What happened?

When decoding tagged maps, they are always returned as `frozendict`, regardless of the value of the `immutable` flag.

I assume this is a bug? Maybe related to #295?

### How can we reproduce the bug?

```
>>> from cbor2 import loads
>>> print(loads(bytes.fromhex("CAA10000")))
CBORTag(10, frozendict({0: 0}))
>>> print(loads(bytes.fromhex("A10000")))
{0: 0}
```

agronholm:
No, this one was a conscious decision. What problems is this causing for you?

oyvindronningstad:
@agronholm Personally, I'm having 3 different issues:

1. I'm traversing the decoded structure and detecting maps via their type (`dict`). This is quite trivial to fix.
2. I'm decoding data, changing it, and reencoding, and frozendicts cannot be modified.
3. I'm encoding the decoding result (or a subset of it) to JSON and the json library doesn't accept frozendict (at least not cbor2.frozendict)

The last two are fixable by taking all results of cbor2 decoding and passing it through a recursive function that rebuilds it with dicts and lists instead of frozendicts and tuples.

Can I ask why this was done? I cannot really see any reason why it shouldn't be controllable. My thinking is that, in principle, there's no difference between cbor data by itself, and cbor data wrapped in a bstr. A protocol could easily include a structure that is sent bare in certain cases, and wrapped in a bstr in others (as part of a bigger structure), and handling these will need to be different with the current change. (unless you resolve to use only frozendicts).

agronholm:
I truly wish you had tried the RC version. It was out for a month but hardly anyone seemed to try it, and the day I shipped 6.0.0 final I get a bunch of issues reported. Now I can't switch the default without breaking backwards compatibility before 7.0.

The reasoning was that immutable structures can be more efficient and tags tend to just contain static data not intended to be mutable anyways.

oyvindronningstad:
Can you add a separate `immutable` flag for tags and bstrs that is default `True`? It's quite restrictive that it's not controllable in any way, and if there are significant performance reasons, maybe the flag should be separate anyways? Just an idea

agronholm:
[...]

PR #312, opened 2026-06-05
Decode unknown-tag content according to the immutable flag

theeldermillenial:
# PR 2 — Decode unknown-tag content according to the `immutable` flag

**Branch:** `fix/tag-content-immutable`

## Problem
The content of a semantic tag with **no designated decoder** is always decoded as immutable — `tuple` for arrays, `frozendict` for maps — regardless of the decoder's `immutable` flag.

Every other container honours `immutable`: top-level arrays/maps, nested arrays/maps, and even the tag-258 (set) decoder all thread the flag through. Only the generic-tag arm in `decode_semantic` hardcodes `true`:

```rust
return Ok(BeginFrame(
    callback,
    true,            // <-- ignores `immutable`
    Some(container),
    DisplayName::SemanticTag(tagnum),
));
```

So `cbor2.loads(cbor2.dumps(CBORTag(6000, [1, 2, 3]))).value` is a `tuple`, even though the caller asked for (default) mutable decoding.

This:
- surprises callers who decode with `immutable=False` and then try to mutate a tag's array/map value;
- breaks downstream libraries that assume `cbor2.loads(tagged).value` is a `list` — the behavior of cbor2 ≤ 5.x. (Concretely, this is what breaks `pycose`'s `CoseMessage.decode`, which does `isinstance(value, list)` and `value.pop(0)`.)

It is a further instance of the mutability bug fixed in 6.0.1 (#295, "values being decoded as immutable in unexpected places").

## Solution
Thread the `immutable` flag through the generic-tag `BeginFrame` so tag content honours it like every other container (`true` → `immutable`).

## Changes
- `rust/decoder.rs`: one-line fix in `decode_semantic`'s generic-tag arm.
- `tests/test_decoder.py`: `test_unknown_tag_value_respects_immutable` — array content is `list` by default / `tuple` under `immutable=True`; map content is `dict` / non-`dict` mapping respectively.
- changelog entry.

## Verification
Full test suite passes (442); no existing test depended on tag content being immutable.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,13 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed the content of a semantic tag with no designated decoder being decoded as immutable
+  (``tuple``/``frozendict``) regardless of the ``immutable`` flag. The nested array/map of such a
+  tag now honours ``immutable`` like every other container. This is a further instance of the
+  mutability bug fixed in 6.0.1 (`#295 <https://github.com/agronholm/cbor2/issues/295>`_).
+
 **6.1.2** (2026-06-02)
 
 - Fixed incorrect tracking of string references for definite-length text strings of length greater
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,12 +5,15 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- **BACKWARD INCOMPATIBLE** Changed the content of a semantic tag to be decoded to adhere to the
+  current "immutable" status of the decoder stack (i.e. when decoding a map key, immutable would
+  be set to ``true``) instead of always decoding as immutable
+  (`#295 <https://github.com/agronholm/cbor2/issues/295>`_)
+
 **6.1.3** (2026-07-04)
 
-- Fixed the content of a semantic tag with no designated decoder being decoded as immutable
-  (``tuple``/``frozendict``) regardless of the ``immutable`` flag. The nested array/map of such a
-  tag now honours ``immutable`` like every other container. This is a further instance of the
-  mutability bug fixed in 6.0.1 (`#295 <https://github.com/agronholm/cbor2/issues/295>`_).
 - Fixed the decoder registering 6-byte strings in the string reference namespace at indices
   65536–4294967295 where the encoder does not, desynchronising the namespace and resolving later
   string references to the wrong value
--- a/rust/decoder.rs
+++ b/rust/decoder.rs
@@ -922,7 +922,7 @@ impl CBORDecoder {
                 });
                 return Ok(BeginFrame(
                     callback,
-                    true,
+                    immutable,
                     Some(container),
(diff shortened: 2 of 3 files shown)


theeldermillenial:
@agronholm I see you've been merging in other PRs.

Is there something wrong with this and #311 that I should fix? Or is the content just something you need to think through?

agronholm:
Those were straightforward bug fixes. This changes the decoding semantics by changing what type gets passed to the decoder hook. This is a potentially backwards incompatible change.

theeldermillenial:
I understand that, but can you suggest a path forward? Are you considering merging this? Is there something I can do that would be better?

At the end of the day the current encoding/decoding loop isn't idempotent. Roundtrip serialization can produce different results, and that's a problem when you're trying to hash encoded cbor to check for similarity.

I'm very dedicated to trying to make this work. Currently we forked the old repo because there were too many issues, but I'm trying to bring us back into your new Rust version. I'd prefer to stay on your version rather than create a new fork so that we have a bigger community and we can all work together on this.

I would like to have some kind of solution in the next few weeks, and I have time to dedicate to implementing this however we need to in order to move forward. As far as I can tell, this is going to require some kind of breaking change.

agronholm:
#296 reported the same issue, yes? If so, that would probably be enough to convince me to make the backwards incompatible change. It will require a major version bump but that's okay.

docs/versionhistory.rst
```diff
@@ -7,6 +7,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed the content of a semantic tag with no designated decoder being decoded as immutable
```
agronholm: First of all, this is not a bug, but a conscious choice. The rationale was that immutable containers would work for a broader set of use cases since they're hashable. But I suppose that was a misguided decision. Nevertheless, the changelog entry needs to be reworded as a behavior change, plus it needs the `**BACKWARD INCOMPATIBLE**` prefix.

docs/versionhistory.rst
```diff
@@ -7,6 +7,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
...
+  (``tuple``/``frozendict``) regardless of the ``immutable`` flag. The nested array/map of such a
+  tag now honours ``immutable`` like every other container. This is a further instance of the
+  mutability bug fixed in 6.0.1 (`#295 <https://github.com/agronholm/cbor2/issues/295>`_).
```
agronholm: This has nothing to do with the mutability bug fixed in v6.0.1.

theeldermillenial:
Yes, that works for me!

We originally forked and modified, but I never like doing that. It splits work and I think makes a less useful tool in the long run.

I also see in that conversation your frustration about not checking out the release candidate. I apologize I did not see that release candidate. I know I did open a PR when you were doing your Rust rewrite that got closed, but I never came back to check it out.

Let me know how to proceed.

theeldermillenial:
@agronholm What are the next steps for this?

agronholm:
I will handle all available PRs, bake them into a new 6.x release and then merge yours.

agronholm (changes requested):
I'd like some wording changes in the changelog.

agronholm:
Okay, this will be in the 7.0 release, whenever it's ready. I'm looking into also including the streaming decoder in that one.

PR #313, opened 2026-06-05
fix string reference namespace threshold for large indices

sahvx655-wq:
The string-reference namespace logic in `decode` decides whether a decoded string is registered for later tag 25 references using a length threshold keyed on the namespace index. While comparing it against the encoder's `maybe_stringref` I noticed the two tables disagree for indices in the 65536..=4294967295 range: the decoder registers strings of length 6 and up, the encoder only length 7 and up. A reference at those indices is tag 25 (two bytes) plus a four-byte uint, so seven bytes total, and the encoder rightly refuses to reference a six-byte string because that would grow the payload. The decoder registering it adds an entry the encoder never created, which shifts every following index by one and makes subsequent references resolve to the wrong value.

Bumping the decoder constant to 7 brings it in line with the encoder and the stringref spec. The rule has to live here because the namespace is rebuilt entirely from the decode side, so it must mirror the encoder exactly or the indices desync silently with no error raised. I have added a regression test that fills the namespace to index 65536 followed by a six-byte and a seven-byte string and checks that a reference resolves to the seven-byte one; on the unpatched tree it comes back as the six-byte string instead.


agronholm:
Thanks.

PR #314, opened 2026-06-08
measure text strings by byte length in maybe_stringref

sahvx655-wq:
**String reference threshold counts code points, not encoded bytes**

While following up on the namespace fix in #313 I lined `maybe_stringref` up against the decoder and found a second place they disagree that the threshold table does not cover: the encoder decides whether to register a string with `value.len()`, which for a `str` is the number of code points, while the length written on the wire and the decoder both count UTF-8 bytes. For any non-ASCII text the two sides then make opposite decisions about crossing the 3/4/5/7/11-byte threshold, the namespace indices drift apart, and subsequent tag 25 references resolve to the wrong earlier string, so the payload comes back silently corrupted instead of erroring. The text branch now measures UTF-8 byte length to match the decoder and the encoded form; bytestrings were already counted in bytes. The added regression round-trips a non-ASCII payload that previously decoded incorrectly.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,10 +7,6 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
-- Fixed the encoder measuring text strings by code point count instead of UTF-8 byte length when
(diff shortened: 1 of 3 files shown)


agronholm (approved):
Nice!

Issue #264, opened 2025-10-21
[Bug]: Definite String Decoder Requests Unlimited Bytes After UTF-8 Chunk Boundary


tylzh97:
### Things to check first

- [x] I have searched the existing issues and didn't find my bug already reported there

- [x] I have checked that my bug is still present in the latest release

### cbor2 version

5.7.0

### Python version

3.10.18

## Summary
`decode_definite_long_string()` keeps `buffer_size` at the old allocation even after UTF-8 leftovers are consumed. The next loop iteration computes `chunk_length = 65536 - buffer_size`, which drops below zero and gets passed straight to `read(-1)`.
Any stream decoder can be driven into an unlimited read followed by `CBORDecodeEOF`.

## poc
```python
import io
import cbor2
import _cbor2

class LoggingReader(io.BytesIO):
    def read(self, n=-1):
        print(f"read({n})")
        return super().read(n)

payload = "a"*65535 + "€" + "b"*65533 + "€" + "d"*100

decoder = _cbor2.CBORDecoder(LoggingReader(cbor2.dumps(payload)))
decoder.decode()
```

## Traceback
```
read(1)
read(4)
read(65536)
read(65536)
read(-1)
Traceback (most recent call last):
  ...
_cbor2.CBORDecodeEOF: premature end of stream (expected to read -1 bytes, got 102 instead)
```

tylzh97:
> Sounds bad. Are you able to make a PR to fix this? I have precious little time allocated to this project.

PR in #265 , also fixed the Github CI Version

PR #316, opened 2026-06-09
avoid quadratic concatenation when decoding chunked strings

sahvx655-wq:
## Changes

Decoding an indefinite-length byte or text string assembles the result by concatenating each chunk onto the running value with `+` (`bytes.add`/`string.add` in `decode_bytestring`/`decode_string`). Every `+` allocates a fresh object and copies the whole accumulator, so a string delivered as N chunks costs O(N^2). I ran into it while profiling a decode of a payload made of many tiny indefinite chunks: a 320 KiB input of single-byte chunks sat at roughly 680 ms, and the time quadruples each time the chunk count doubles. The same construct is used on the large definite-length (>65536 byte) paths. As the decoder runs over untrusted input with the default settings, a small crafted payload of many one-byte chunks turns into seconds of CPU and a lot of short-lived allocation, so the practical effect is a cheap denial of service rather than merely a slow path.

The original pure-Python decoder gathered the chunks in a list and did a single `b"".join`/`"".join` at the end, which is linear, and I have restored that shape here. Byte chunks accumulate into a `Vec<u8>` that becomes one `bytes` object, and text chunks accumulate into a list joined once, with the join kept in Python so the `str_errors` surrogate handling is untouched. The same input now decodes in about 9 ms and scales linearly. Keeping the assembly at the point where the chunks are read means every string path benefits without any caller change.

Before/after on indefinite byte strings of single-byte chunks:

```
chunks    input     before      after
 40000     80 KiB    21.6 ms     2.4 ms
 80000    160 KiB   130.3 ms     5.0 ms
160000    320 KiB   682.4 ms     9.3 ms
```

## Checklist

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,10 +7,6 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
-- Fixed quadratic decoding time for indefinite-length and large definite-length byte and text
-  strings, caused by concatenating each chunk onto the accumulated result with ``+`` instead of
-  building the result once
-  (`#316 <https://github.com/agronholm/cbor2/pull/316>`_; PR by @sahvx655-wq)
 - Fixed the decoder registering 6-byte strings in the string reference namespace at indices
   65536–4294967295 where the encoder does not, desynchronising the namespace and resolving later
   string references to the wrong value
@@ -18,6 +14,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed the IPv4/IPv6 network decoders (tags 52 and 54) silently truncating an address byte string
   that is longer than the address size instead of rejecting it as malformed
   (`#309 <https://github.com/agronholm/cbor2/pull/309>`_; PR by @sahvx655-wq)
+- Fixed quadratic decoding time for indefinite-length and large definite-length byte and text
+  strings, caused by concatenating each chunk onto the accumulated result with ``+`` instead of
+  building the result once
+  (`#316 <https://github.com/agronholm/cbor2/pull/316>`_; PR by @sahvx655-wq)
 
 **6.1.2** (2026-06-02)
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed quadratic decoding time for indefinite-length and large definite-length byte and text
+  strings, caused by concatenating each chunk onto the accumulated result with ``+`` instead of
+  building the result once
+  (`#316 <https://github.com/agronholm/cbor2/pull/316>`_; PR by @sahvx655-wq)
 - Fixed the decoder registering 6-byte strings in the string reference namespace at indices
   65536–4294967295 where the encoder does not, desynchronising the namespace and resolving later
   string references to the wrong value
--- a/rust/decoder.rs
+++ b/rust/decoder.rs
@@ -428,9 +428,9 @@ impl CBORDecoder {
             Some(length) => {
                 // Incrementally read the bytestring, in chunks of 65536 bytes. The claimed
                 // length is untrusted until the data has actually been read, so no more than
-                // 1 MiB of it is reserved up front; a truncated payload claiming a huge length
-                // can then force at most a 1 MiB allocation.
-                let bytes = PyBytes::new_with_writer(py, length.min(1_048_576), |writer| {
+                // 64 KiB of it is reserved up front; a truncated payload claiming a huge length
+                // can then force at most a 64 KiB allocation.
+                let bytes = PyBytes::new_with_writer(py, length.min(65536), |writer| {
                     let mut remaining_length = length;
                     while remaining_length > 0 {
                         let chunk_size = remaining_length.min(65536);
--- a/rust/decoder.rs
+++ b/rust/decoder.rs
@@ -21,7 +21,6 @@ use pyo3::types::{
 };
 use pyo3::{IntoPyObjectExt, Py, PyAny, PyErrArguments, intern, pyclass};
 use std::fmt::{Display, Formatter};
-use std::io::Write;
 use std::mem::{replace, take};
 
 const IMMUTABLE_ATTR: &str = "_cbor2_immutable";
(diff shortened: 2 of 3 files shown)


agronholm:
Have you considered an approach involving [PyBytes.new_with_writer](https://docs.rs/pyo3/latest/pyo3/types/struct.PyBytes.html#method.new_with_writer)? That would potentially avoid having to keep the contents in memory twice (even if it's brief).

sahvx655-wq:
I hadn't, and it's a better fit — switched both bytestring paths over to `PyBytes::new_with_writer` so the contents only ever live in the buffer the writer is growing, with no transient `Vec` copy at the end. Before doing it I read through the writer implementation to make sure the linear behaviour survives: each `write_all` goes through a resize call, and if that resized to the exact requested size every time we'd be straight back to quadratic. It doesn't — the compat shim over-allocates by 25% (50% on Windows) and the native 3.15 `PyBytesWriter` over-allocates similarly, so growth is geometric and the whole thing stays amortised linear. Re-ran the 400k single-byte chunk payload to confirm: ~65 ms, scaling linearly per doubling. That's a somewhat higher constant than the `Vec` version (~22 ms, since `extend_from_slice` is cheaper than an FFI resize per chunk) but it's nowhere near the 13.8 s of the old code, and not holding a second full copy of a large payload is the better trade.

One deliberate choice worth flagging: on the definite-length path I pass 0 as the reserved capacity rather than reserving the claimed length up front. The length field is attacker-controlled until the data has actually been read, and pre-allocating it would let a short truncated payload claiming a multi-gigabyte string force a huge allocation — which is exactly what the existing 65536-byte incremental reads are there to prevent. I left the text string paths on the single `join` since there's no equivalent writer for `str` and the chunks have to pass through the UTF-8 decoder anyway. Full test suite passes.

agronholm:
> One deliberate choice worth flagging: on the definite-length path I pass 0 as the reserved capacity rather than reserving the claimed length up front. The length field is attacker-controlled until the data has actually been read, and pre-allocating it would let a short truncated payload claiming a multi-gigabyte string force a huge allocation — which is exactly what the existing 65536-byte incremental reads are there to prevent. I left the text string paths on the single `join` since there's no equivalent writer for `str` and the chunks have to pass through the UTF-8 decoder anyway. Full test suite passes.

Good call. IIRC `cbor2` had a related vulnerability in the past.

agronholm:
The test suite is failing due to a linter error.

agronholm:
> One deliberate choice worth flagging: on the definite-length path I pass 0 as the reserved capacity rather than reserving the claimed length up front. The length field is attacker-controlled until the data has actually been read, and pre-allocating it would let a short truncated payload claiming a multi-gigabyte string force a huge allocation — which is exactly what the existing 65536-byte incremental reads are there to prevent. I left the text string paths on the single `join` since there's no equivalent writer for `str` and the chunks have to pass through the UTF-8 decoder anyway. Full test suite passes.

Wouldn't it be okay to reserve the indicated length if it's reasonably small? Or will this have a negligible impact on performance anyway?

sahvx655-wq:
The lint failure was the `std::io::Write` import I'd added: `new_with_writer` hands the closure a `&mut dyn Write`, and method calls on a trait object resolve without the trait being in scope, so rustc flagged the import as unused and the workflow's `RUSTFLAGS: --deny warnings` promoted that to a build error. I'd been building locally without the deny flag, which is why it only surfaced in the CI log. Import dropped; `cargo check`, `clippy` and `fmt` are all clean under `--deny warnings` now.

On reserving: it's measurably worth it, so the definite-length path now reserves `length.min(1_048_576)`. Anything up to 1 MiB gets a single exact allocation with no resize copies, larger strings still get their first 1 MiB free, and the worst a truncated payload claiming a huge length can force is a transient 1 MiB allocation, the same order as the 64 KiB chunk buffers the read loop already churns through. Benchmarked on release builds (best of 7): a 128 KiB definite bytestring decodes in 0.016 ms vs 0.036 ms without the reservation, 1 MiB in 0.12 ms vs 0.30 ms, 16 MiB in 3.3 ms vs 4.0 ms, and by 128 MiB the two converge at roughly 16 ms since geometric growth dominates either way. So not negligible in the 64 KiB to 16 MiB range, roughly 1.5 to 2x. Full suite passes (446) and the 400k-chunk indefinite payload still decodes in about 11 ms, so the linear behaviour is unchanged.

agronholm:
If it allocates 1 MB by default, I feel that leaves potential for DoS attacks. How about 64 KB instead? Most strings are not that large anyway.

sahvx655-wq:
Agreed, that 1 MiB ceiling was generous and a worst-case payload could fan that out across many small strings. Dropped it to 64 KiB, which also lines up neatly with the 65536-byte chunk size the read loop already uses, so a truncated payload claiming a huge length now forces at most one chunk-sized allocation.

The performance trade is fine: strings up to 64 KiB still get their single exact allocation, and anything larger only loses the first reservation before geometric growth takes over, so the gain over the no-reserve version is marginal past that point anyway. Re-ran the 400k single-byte chunk payload (linear, ~55 ms) and the full suite still passes.

agronholm:
I'm satisfied with the way `bytes` objects are created now, but the string creation part looks suboptimal. Since we're using twice the memory required to hold the actual string anyway, why not use `PyBytes.new_with_writer` here too, and then pass the resulting `bytes` object to `PyString::from_encoded_object()`? Is there something I'm not considering here?

sahvx655-wq:
Done for the definite-length path. It now reads the chunks straight into a `PyBytes::new_with_writer` buffer, with the same `length.min(65536)` reservation as the bytestring path so a truncated payload claiming a huge length still can't force more than a 64 KiB allocation, then hands the finished bytes to `PyString::from_encoded_object` with `str_errors` mapped onto the errors argument. That also let me drop the incremental UTF-8 decoder and its `PyOnceLock` altogether, since decoding the assembled buffer in one pass handles a multi-byte character split across a 64 KiB read boundary on its own, which was the only reason the incremental decoder was there (the #264 test still passes). Re-ran the suite plus a few large strings and decode stays linear, about 70 ms for 64 MiB.

The one place I didn't apply it is the indefinite-length path, and that's deliberate. There each chunk is its own definite-length text string, and RFC 8949 says a single UTF-8 character can't be spread across chunks, so a chunk ending mid-character is malformed. The current per-chunk decode rejects that (or applies str_errors per chunk), whereas concatenating the raw bytes first and decoding the lot in one go would silently stitch a split character back together and accept input the spec forbids. So I kept that path on decode-then-join to preserve the boundary semantics.

agronholm (approved):
Looks good now, thanks!

PR #320, opened 2026-06-15
Reject CBOR break codes outside indefinite-length items

### LINKED ISSUE #305: Confusing decoding of standalone break marker; `allow_indefinite=False` still accepts it
### Things to check first

### cbor2 version

6.0.1

### Python version

3.11.15

### What happened?

```python
import sys
import cbor2

CASES = [
]

    try:
        return True
    return False

print()

ok = True

LONE = b"\xff"
print()

sys.exit(0 if ok else 1)
```

```
   cbor2 version: 6.0.1
   python:        3.11.15

```

binggao1230:
Fixes #305.

## Problem

A break code (major type 7, subtype 31 / `0xff`) is only well-formed as the terminator of an indefinite-length item. When it appears anywhere a data item is expected, the decoder leaks the internal break-marker sentinel to the caller (or accepts it as a real element) instead of rejecting the input:

```python
>>> import cbor2
>>> cbor2.loads(bytes.fromhex('ff'))                          # naked break
<object object at 0x...>                                      # should raise
>>> cbor2.loads(bytes.fromhex('ff'), allow_indefinite=False)
<object object at 0x...>                                      # should raise
>>> cbor2.loads(bytes.fromhex('8301ff02'))                    # break inside a definite array
[1, <object object at 0x...>, 2]                              # should raise
```

The same leak occurs for definite-length map keys/values (`a1ff01`, `a101ff`) and for the content of a tag, shareable value or string-reference namespace (`d818ff`, `d81cff`, `d90100ff`, `d9d9f7ff`). RFC 8949 sect. 3.2.1 defines such input as not well-formed, and as noted in the issue, *"under no circumstances should a naked break marker come out of the decoding process"*.

## Cause

The break marker is produced by `decode_special` as an ordinary decoded value and flows through the generic frame machinery. Only the indefinite-length array/map callbacks check `item.is(break_marker)` to terminate; every other consumer (definite-length containers, tags, sets, shareables, string namespaces, and the top-level result) silently passes it through.

## Fix

Establish the invariant that a break marker may only be consumed by an indefinite-length container. A new `accepts_break` flag on `StackFrame` is set only for indefinite-length array and map frames (indefinite-length strings consume the break internally and never produce a free value). The decode loop then rejects a break marker that reaches any other frame, and one left as the top-level result, with a `CBORDecodeError`.

## Tests

Added `test_break_outside_indefinite` (12 positions: naked, before-data, definite array element/first/nested, definite map key/value, tag content, shareable, string namespace, self-describe tag, set), `test_break_outside_indefinite_disabled`, and `test_break_terminates_indefinite` (5 cases confirming real indefinite arrays/maps, nesting, and indefinite-in-definite still decode). `test_string_oversized`'s fuzz payload embeds a break inside a definite map and now raises the break error before its truncated tail is reached, so its expectation was updated (premature-end-of-stream remains covered by other tests). Full suite passes (462).

I worked on this with AI assistance under my direction and reviewed the change myself.

### DIFF
--- a/rust/decoder.rs
+++ b/rust/decoder.rs
@@ -47,9 +47,10 @@ static FROZEN_DICT: PyImportable = PyImportable::new("builtins", "frozendict");
 enum DecoderResult<'a> {
     BeginFrame(
         Box<DecoderCallback<'a>>,
-        bool,
+        bool, // requested_immutable
         Option<Bound<'a, PyAny>>,
         DisplayName<'a>,
+        bool, // accepts_break: an indefinite-length container that a break code terminates
     ),
     ContinueFrame(bool),
     CompleteFrame(Bound<'a, PyAny>),
@@ -86,6 +87,9 @@ struct StackFrame<'py> {
     shareable_index: Option<usize>,
     typename: DisplayName<'py>,
     contains_string_namespace: bool,
+    // Whether a CBOR break code is a legal terminator here (only indefinite-length
+    // arrays and maps). A break reaching any other frame is ill-formed input.
+    accepts_break: bool,
 }
 
 /// Decorates a function to be a two-stage decoder.
@@ -565,6 +569,7 @@ impl CBORDecoder {
                 false,
                 None,
                 DisplayName::String("array"),
+                optional_length.is_none(),
             ))
         } else {
             let mut list = PyList::empty(py);
@@ -602,6 +607,7 @@ impl CBORDecoder {
                 false,
                 Some(container),
                 DisplayName::String("array"),
+                optional_length.is_none(),
             ))
         }
     }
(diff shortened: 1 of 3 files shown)


tests/test_decoder.py
```diff
@@ -396,7 +396,10 @@ def test_string_invalid_utf8(payload: str) -> None:
 
 
 def test_string_oversized() -> None:
-    with pytest.raises(CBORDecodeEOF, match="premature end of stream"):
+    # This payload embeds a break code (0xff) inside a definite-length map, which is
+    # now rejected as ill-formed before the truncated tail is reached (premature end
+    # of stream is exercised separately).
+    with pytest.raises(CBORDecodeError, match="break code"):
         loads(unhexlify("aeaeaeaeaeaeaeaeae0108c29843d90100d8249f0000aeaeffc26ca799"))
```
agronholm, line 411: Seems to me this is covered by the test you added, so this whole test can be deleted

rust/decoder.rs
```diff
@@ -1591,6 +1609,15 @@ impl CBORDecoder {
             let result: PyResult<DecoderResult<'py>> = if let Some(previous_value) = value.take() {
                 // Call the decoder callback of the last frame
                 let frame = frames.last_mut().unwrap();
+                // A break code is only a valid terminator inside an indefinite-length
```
agronholm, line 1667:
```suggestion

                // A break code is only a valid terminator inside an indefinite-length
```

PR #321, opened 2026-06-18
fix: respect immutable flag for unknown semantic tags

### LINKED ISSUE #296: Tag values always decoded as immutable

gaoflow:
When a CBOR tag wraps a map, the decoder hardcoded
`requested_immutable=true` when creating the BeginFrame for the tag's
inner content. This forced the inner map to always decode as
`frozendict` regardless of the `immutable` argument passed to the
decoder.

Change `true` to the `immutable` parameter from `decode_semantic`,
so the user's preference is respected:
- `immutable=False` (default) → regular `dict`
- `immutable=True` → `frozendict`

Fixes: #296

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed the decoder leaking the internal break-marker sentinel (or accepting it as a data item)
+  when a CBOR break code (``0xff``) appears outside an indefinite-length item, instead of rejecting
+  such input as ill-formed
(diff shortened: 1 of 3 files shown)


agronholm:
This looks like an AI-generated slop PR.

agronholm:
Oh, and besides, it's a duplicate of #312.

PR #327, opened 2026-07-14
Decode self-describe and unhandled tag containers as mutable

sahvx655-wq:
## Changes

The self-describe tag (55799) arm in `decode_semantic` opens its content with a frame that forces immutable decoding, so a map or array behind the tag comes back as a `frozendict` or `tuple` where the reference pure-Python decoder returns a `dict` or `list`. I found it while diffing decode output against the old pure-Python decoder; the visible effect is a surprise `TypeError` on decode-then-mutate, or a failing `type(x) is list` check, but only when a producer prepends the self-describe magic number.

The tag is transparent per RFC 8949, so the arm now returns its own frame inheriting the ambient immutability, the same way the unknown-tag arm does on this branch since #312. A tag in an immutable position such as a map key still yields a hashable value, and the tuple-consuming tags (fraction, bigfloat, rational, complex, the IPv4/IPv6 array forms) keep their forced-immutable path untouched.

## Checklist

### DIFF
--- a/rust/decoder.rs
+++ b/rust/decoder.rs
@@ -915,10 +915,17 @@ impl CBORDecoder {
             261 => (Box::new(Self::decode_ipnetwork), "IP network"),
             1004 => (Box::new(Self::decode_date_string), "string-form date"),
             43000 => (Box::new(Self::decode_complex), "complex number"),
-            55799 => (
-                Box::new(Self::decode_self_describe_cbor),
-                "self-described CBOR value",
-            ),
+            55799 => {
+                // The self-describe tag is transparent: its content must decode exactly as it
+                // would without the tag, so it inherits the ambient immutability instead of
+                // forcing the value immutable like the value-consuming tags above.
+                return Ok(BeginFrame(
+                    Box::new(Self::decode_self_describe_cbor),
+                    immutable,
+                    None,
+                    DisplayName::String("self-described CBOR value"),
+                ));
+            }
             _ => {
                 // For a tag with no designated decoder, check if we have a tag hook, and call
                 // that with the tag object, using its return value as the decoded value.
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -11,6 +11,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   current "immutable" status of the decoder stack (i.e. when decoding a map key, immutable would
   be set to ``true``) instead of always decoding as immutable
   (`#295 <https://github.com/agronholm/cbor2/issues/295>`_)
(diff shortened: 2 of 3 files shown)


agronholm:
This was already done for unhandled tag containers via #312. Does this PR add anything new?

sahvx655-wq:
I re-ran current master (18c6507) before replying, and both cases still decode immutable, not just the self-describe one:

```
self-describe map -> frozendict
self-describe arr -> tuple
unhandled map     -> frozendict
unhandled arr     -> tuple
```

The unhandled-tag arm in `decode_semantic` still hard-codes `true` on its `BeginFrame`, so it looks like the `true` -> `immutable` change from #312 isn't in master at the moment (reverted or lost in a later merge?). That's why the map/array behind an unhandled tag still comes back frozen.

The self-describe tag (55799) is a separate arm from that fallback and #312 never touched it, so it stays broken regardless. This PR fixes that arm and also restores the unhandled-tag behaviour that's currently missing. A tag used in an immutable position such as a map key still yields a hashable `frozendict`, so nothing changes there. If you've already re-landed #312's half somewhere and only want the self-describe change, I'm happy to drop the fallback hunk.

agronholm:
The linked PR was merged into the 7.0 branch which is the code for the upcoming 7.0 major release. That PR (and this one) introduces a backwards-incompatible change which is why it was done like that. If this PR adds value, then it should also be targeted at the 7.0 branch.

sahvx655-wq:
That explains it; I was diffing against master, which is why the unhandled-tag case looked broken to me as well. Makes sense to keep behaviour changes like this out of 6.x.

Retargeted at 7.0 and rebased. With #312 already on that branch the fallback hunk was redundant, so I dropped it; the PR is now just the self-describe (55799) arm, which #312 didn't touch and which still forces its content immutable there. Without it, anything prefixed with the self-describe magic number still round-trips to frozen containers on 7.0. Full suite passes on the rebased branch (453 tests).

rust/decoder.rs
```diff
@@ -915,10 +915,17 @@ impl CBORDecoder {
...
+                // The self-describe tag is transparent: its content must decode exactly as it
+                // would without the tag, so it inherits the ambient immutability instead of
+                // forcing the value immutable like the value-consuming tags above.
+                return Ok(BeginFrame(
+                    Box::new(Self::decode_self_describe_cbor),
+                    immutable,
+                    None,
+                    DisplayName::String("self-described CBOR value"),
+                ));
+            }
```
agronholm, line 950: Where is the corresponding `CompleteFrame` for this `BeginFrame`?

agronholm: Never mind, I was confused.

agronholm (approved):

sahvx655-wq:
Thanks for merging this, glad the 7.0 branch turned out to be the right home for it.

PR #329, opened 2026-07-15
Encode subclasses of the specially handled types instead of rejecting them

binggao1230:
## Changes

Since the Rust rewrite, encoding a subclass of any of the specially handled types raises `CBOREncodeError`:

```python
>>> import cbor2, datetime
>>> class MyDatetime(datetime.datetime): pass
>>> cbor2.dumps(MyDatetime(2020, 1, 1, tzinfo=datetime.timezone.utc))
cbor2.CBOREncodeError: cannot encode type <class '__main__.MyDatetime'>
```

cbor2 5.x encoded these like their base type (verified on 5.9.0, both the C extension and the pure-Python implementation — `_find_encoder` used `issubclass()`), so things like `pandas.Timestamp` or `Decimal`/`UUID` subclasses used to work. It's also inconsistent with the primitives, whose subclasses (`int`, `str`, `bytes`, ...) still encode fine in 6.x. The 6.0 changelog doesn't list this among the backward-incompatible changes, so I'm assuming it's unintended.

The fix keeps the exact-identity pass over the stdlib encoder table first, so the common case still only pays pointer comparisons, and adds an `isinstance()` fallback pass for subclasses. The table is reordered so subclasses precede their bases in that fallback (`IPv4Interface`/`IPv6Interface` before the address types, like `datetime` before `date`) — so an interface subclass is encoded as an interface, not silently downgraded to a bare address the way 5.x did it.

## Checklist

features

### DIFF
--- a/rust/encoder.rs
+++ b/rust/encoder.rs
@@ -583,6 +583,13 @@ impl CBOREncoder {
                         ),
                         (
                             py.import("ipaddress")?
+                                .getattr("IPv4Interface")?
+                                .cast_into()?
+                                .unbind(),
+                            CBOREncoder::encode_ipv4_interface,
+                        ),
+                        (
+                            py.import("ipaddress")?
                                 .getattr("IPv4Address")?
                                 .cast_into()?
                                 .unbind(),
@@ -597,10 +604,10 @@ impl CBOREncoder {
                         ),
                         (
                             py.import("ipaddress")?
-                                .getattr("IPv4Interface")?
+                                .getattr("IPv6Interface")?
                                 .cast_into()?
                                 .unbind(),
-                            CBOREncoder::encode_ipv4_interface,
+                            CBOREncoder::encode_ipv6_interface,
                         ),
                         (
                             py.import("ipaddress")?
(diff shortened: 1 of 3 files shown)


rust/encoder.rs
```diff
@@ -581,6 +581,13 @@ impl CBOREncoder {
                             py.import("re")?.getattr("Pattern")?.cast_into()?.unbind(),
                             CBOREncoder::encode_regexp,
                         ),
+                        (
```
agronholm, line 584: Why did you move these around?

rust/encoder.rs
```diff
@@ -632,12 +632,22 @@ impl CBOREncoder {
                         ),
                     ])
                 })?;
+            // Exact type matches first: the common case, using cheap pointer comparisons
```
agronholm, line 635:
```suggestion

            // Exact type matches first: the common case, using cheap pointer comparisons
```

tests/test_encoder.py
```diff
@@ -436,6 +436,116 @@ def test_ipaddress(value: object, expected: str) -> None:
...
+class _IPv6NetworkSubclass(IPv6Network):
+    pass
+
+
+class _IPv6InterfaceSubclass(IPv6Interface):
+    pass
+
+
+class _MIMETextSubclass(MIMEText):
+    pass
```
agronholm, line 484: This seems rather excessive.

agronholm (changes requested):
Overall looks good. I could add this in v7.0.

PR #334, opened 2026-08-15
Hash CBORSimpleValue consistently with its integer value

sahvx655-wq:
1. `__richcmp__` treats `CBORSimpleValue(n)` as equal to the integer `n`, but `__hash__` ran the raw byte through a Rust `DefaultHasher`, so equal values hashed differently and broke Python's rule that equal objects hash alike.
2. As a result a decoded simple value could not be found by its integer in a dict or set, and under `allow_duplicate_keys=False` a map carrying both `5` and simple value `5` was accepted rather than rejected as a duplicate, since the two landed in different hash buckets.
3. Restored `__hash__` to return the hash of the underlying integer, matching the comparison semantics and the behaviour of the pure-Python decoder this port replaced.

I noticed this while auditing the `__hash__` implementations in `types.rs`: the comparison operators fall back to comparing against a plain `int`, which only makes sense if the hashes line up too. Regression tests for the key interchangeability and for the duplicate-key check are included.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,10 +7,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
-- Fixed :class:`CBORSimpleValue` hashing inconsistently with the integer it compares equal to, so
-  that ``CBORSimpleValue(n) == n`` held while ``hash(CBORSimpleValue(n)) != hash(n)``. A decoded
-  simple value could not be looked up by its integer in a mapping or set, and an integer key
-  colliding with a simple-value key slipped past the ``allow_duplicate_keys=False`` check
+- Fixed :class:`CBORSimpleValue` hashing inconsistently with the integer it compares equal to
   (`#334 <https://github.com/agronholm/cbor2/pull/334>`_; PR by @sahvx655-wq)
 
 **6.1.4** (2026-08-01)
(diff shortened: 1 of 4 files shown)


docs/versionhistory.rst
```diff
@@ -5,6 +5,14 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed :class:`CBORSimpleValue` hashing inconsistently with the integer it compares equal to, so
+  that ``CBORSimpleValue(n) == n`` held while ``hash(CBORSimpleValue(n)) != hash(n)``. A decoded
+  simple value could not be looked up by its integer in a mapping or set, and an integer key
+  colliding with a simple-value key slipped past the ``allow_duplicate_keys=False`` check
```
agronholm: Cut out the unnecessary blabber.
```suggestion
- Fixed :class:`CBORSimpleValue` hashing inconsistently with the integer it compares equal to
```

agronholm:
Pre-commit checks are failing. Please fix them.

sahvx655-wq:
Fixed. The failure was mypy on the new test: indexing a literal `dict[int, str]` with a `CBORSimpleValue` trips the `[index]` check even though it works at runtime, which is of course the whole point of the test.

Annotated the mapping as `dict[int | CBORSimpleValue, str]` so the lookups type-check in both directions. `pre-commit run --all-files` is clean locally now and the type/decoder tests still pass.

agronholm (approved):

PR #335, opened 2026-09-01
Give a nested stringref namespace its own index space

dylanpulver:
## Changes

`encode_semantic()` (`rust/encoder.rs:915`) turned `string_referencing` on for tag
256 but left the outer `string_references` / `bytes_references` maps in place —
the two `TODO`s in that function marked exactly this. References emitted inside a
nested namespace were therefore numbered against the *outer* namespace, and
strings registered inside it leaked back out. The decoder
(`rust/decoder.rs:1735`) already handles it correctly: it pushes a fresh
namespace and resolves tag 25 against the innermost one.

So the encoder emits output its own decoder cannot read — and sometimes reads
back as different data, with no exception:

```python
>>> v = ["aaa", CBORTag(256, ["bbb", "ccc", "ddd", "bbb"])]
>>> loads(dumps(v, string_referencing=True))
['aaa', ['bbb', 'ccc', 'ddd', 'ccc']]        # 'bbb' came back as 'ccc'
```

The [stringref spec](http://cbor.schmorp.de/stringref) requires the reset:

> Since stringref-namespace tags can be nested, the decoder needs to **save and
> restore the outer array** before starting and after ending the decoding of the
> tagged value.

The fix sets the outer maps aside on entry and restores them on exit.

### Oracle: the spec's own worked example

That page publishes a nested example with its bytes and its decoding. On `master`,
`loads()` reproduces the documented value exactly, but `dumps()` of that same
value produces different bytes, and `loads()` then fails on them:

| | `master` | this PR |
|---|---|---|
| decode spec bytes → documented value | ✅ | ✅ |
| encode documented value → spec bytes | ❌ | ✅ |
| round trip through cbor2's own decoder | ❌ `string reference 2 not found` | ✅ |

The new test asserts against those published bytes rather than against anything
this implementation produced.

### Test runs (same command and environment each time)

`python -m pytest tests -q`

| | result |
|---|---|
| clean `master` | `464 passed` |
| `master` + the new tests | `2 failed, 465 passed` |
| this PR | `467 passed` |

I also built the half-fix — clear the maps on entry, don't restore them — and it
still fails both new tests, which is why the restore case is tested separately.
`test_decode_..._matches_spec_example` passes either way; it is there as a
control showing the decoder was never the faulty side.

Not a security issue as far as I can tell: it is encoder-side, `string_referencing`
is opt-in, and it needs a `CBORTag(256, …)` in the caller's own data. I'll add the
changelog PR link once this has a number.

## Checklist

- [x] You've added a new changelog entry (in `docs/versionhistory.rst`)

---

AI disclosure: found and drafted with Claude Code (model Claude Opus 5,
`claude-opus-5`) — differential fuzzing of encode→decode round trips, then the
spec's published vector as the oracle. Reviewed before submission.


docs/versionhistory.rst
```diff
@@ -7,6 +7,12 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed the encoder not giving a nested stringref namespace (tag 256) its own index space. The
+  outer namespace was neither set aside on entry nor restored on exit, so string references emitted
+  inside the nested namespace were numbered against the outer one while the decoder resolves them
+  against the inner one; depending on the data this raised ``string reference N not found`` or
+  silently decoded a repeated string as a different string
```
agronholm: Removed excessive AI blabber
```suggestion
- Fixed the encoder not giving a nested stringref namespace (tag 256) its own index space
```

tests/test_encoder.py
```diff
@@ -717,6 +717,52 @@ def test_encode_stringrefs_repeated_bytearray() -> None:
     assert loads(encoded) == [b"abcd", b"abcd", b"abcd"]
 
 
+# The worked nested example published in the stringref spec
+# (http://cbor.schmorp.de/stringref), with its documented decoding.
+SPEC_NESTED_EXAMPLE = unhexlify(
+    "d90100"  # 256(
```
agronholm: These comments need to be horizontally aligned so they make any visual sense.

agronholm: If the linter is preventing that, then use a separate comment block.

agronholm: If that doesn't make sense either, let me know.

agronholm (changes requested):
You just pushed hundreds of generated build files. How about getting rid of those?

PR #337, opened 2026-09-06
Restore built-in exception inheritance

### LINKED ISSUE #299: `cbor2.CBORDecodeError` should be a subclass of `ValueError`
### Things to check first

### cbor2 version

6.0.1

### Python version

3.12

### What happened?

With the recently released `cbor2` 6.0.x, `cbor2.CBORDecodeError` is not a subclass of `ValueError`.

When using data serialization libraries like the stdlib's `json`, it is a common convention that `.loads(...)` or similar raises a subclass of `ValueError` (the passed value was incorrect). This used to work fine in `cbor2` as well until the very recent 6.0.x releases.

Using older `cbor2` (5.x), this would have worked just fine:
```python
try:
    cbor2.loads(b'\xde\xad\xbe\xef')
except ValueError:
    print('Validation failed')
```

On 6.0.x, the decoding error is not caught as `ValueError`.

be-student:
## Changes

Fixes #299.

Restore the built-in exception ancestry exposed by cbor2 5.x while retaining the cbor2 6 exception names:

- `CBOREncodeTypeError` directly inherits from both `CBOREncodeError` and `TypeError`.
- `CBOREncodeValueError` directly inherits from both `CBOREncodeError` and `ValueError`.
- `CBORDecodeError` directly inherits from both `CBORError` and `ValueError`, taking over the role of the removed `CBORDecodeValueError`.
- `CBORDecodeEOF` directly inherits from both `CBORDecodeError` and `EOFError`.

The multi-base exception types are created with the public `pyo3-ffi` `PyErr_NewExceptionWithDoc` API requested in the maintainer feedback. They are cached as Python type objects and used through PyO3's public dynamic `PyErr::from_type` and `PyErr::is_instance` APIs. This avoids custom `PyTypeInfo` implementations and PyO3 internal registration helpers.

The type stub now declares `CBORDecodeError(CBORError, ValueError)`. The runtime test checks the exact direct bases, mypy checks that `CBORDecodeError` is assignable to `ValueError`, and the changelog accurately distinguishes `CBORDecodeError` from the cbor2 5.x `CBORDecodeValueError` class.

## Validation

Validated at `19cb17b3dd42edc1badd3a1bedae7aec38c4abed`:

- Python 3.14.6: `471 passed`
- Focused exception/type tests: `33 passed`
- All configured pre-commit hooks passed, including ruff, mypy, RST checks, and `cargo fmt`
- `cargo clippy --all-targets` passed with `RUSTFLAGS="--deny warnings"` for the Rust implementation; the later commit changed only the stub, changelog, and tests
- Python 3.15.0a8: the native extension built and installed with warnings denied; local runtime import could not be exercised because this preview interpreter lacks the `_PyType_FromSlots` symbol required by PyO3 0.29

## Checklist

- [x] Added tests that fail without the restored exception hierarchy
- [x] Updated the public type stub
- [x] Updated the changelog

## AI disclosure

OpenAI Codex assisted with the implementation, tests, validation, and preparation of this pull request text.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Restored the built-in exception ancestry declared by the public type interface and provided by
+  cbor2 5.x: :exc:`CBORDecodeError` is again a :exc:`ValueError`, encoder type and value errors are
+  again :exc:`TypeError` and :exc:`ValueError` subclasses, and :exc:`CBORDecodeEOF` is again an
+  :exc:`EOFError` (`#299 <https://github.com/agronholm/cbor2/issues/299>`_)
 - Fixed :class:`CBORSimpleValue` hashing inconsistently with the integer it compares equal to
   (`#334 <https://github.com/agronholm/cbor2/pull/334>`_; PR by @sahvx655-wq)
 
--- a/rust/lib.rs
+++ b/rust/lib.rs
@@ -47,16 +47,12 @@ mod _cbor2 {
     #[pymodule_export]
     use crate::types::CBOREncodeError;
 
-    #[pymodule_export]
     use crate::types::CBOREncodeTypeError;
 
-    #[pymodule_export]
     use crate::types::CBOREncodeValueError;
 
-    #[pymodule_export]
     use crate::types::CBORDecodeError;
 
-    #[pymodule_export]
     use crate::types::CBORDecodeEOF;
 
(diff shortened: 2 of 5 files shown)


agronholm:
Seems like a clever solution to the problem, but I'm not sure how kosher or maintainable it is. I'll have to consult the PyO3 folks first.

agronholm:
Also this doesn't compile on Python 3.15, so at least one fix is needed.

agronholm:
I'm told that this is not a good approach, and `pyo3_ffi` should be used instead. I cannot advise further, but given that, I'll need a different approach before I can consider this.

agronholm:
The approach is otherwise palatable except for the use of PyO3's internal API. That I cannot allow.

PR #338, opened 2026-09-06
Add optional mutable byte string decoding

### LINKED ISSUE #266: Add option to decode type bytestring as a mutable bytearray rather than bytes
### Things to check first

- [x] I have searched the existing issues and didn't find my feature already requested there

### Feature description

For performance reasons, it would be nice if there were an option / flag to decode cbor bytes strings (cbor type 2) as python bytearrays rather than bytes. This would allow the user to modify the buffer rather than do a memcpy first to convert it to a mutable type.

This change would actually eliminate two memcpys. The current implementation (I'm looking at decode.c decode_definite_long_bytestring()) copies the packet (in chunks) into a bytearray and then copies the bytearray to a bytes object before returning it. Then to get a mutable array, the user has to copy the bytes into a bytearray.
If the decode just returned the bytearray instead, that would remove two memcpys (bytearray -> bytes during decode, then bytes -> bytearray in python to make it mutable).

From looking through the issues, I think this might affect other use cases:
Issue #15
Issue #240 
So maybe it would be best to make it an optional decode flag?

### Use case

My use case is receiving cbor packets that include large chunks of bytes (images).  My end goal is to get a mutable numpy array using np.frombuffer().

be-student:
## Changes

Fixes #266.

Adds `mutable_bytes=True` to `CBORDecoder`, `load()`, and `loads()` so ordinary CBOR byte strings decode as `bytearray`. Immutable contexts such as map keys and semantic tag payloads remain `bytes`.

## Checklist

- [x] Added regression tests
- [x] Updated documentation
- [x] Added a changelog entry

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -23,6 +23,19 @@ Serializing and deserializing with cbor2 is pretty straightforward::
 
 Some data types, however, require extra considerations, as detailed below.
 
+Mutable byte strings
+--------------------
+
+By default, CBOR byte strings are decoded as :class:`bytes`. Pass ``mutable_bytes=True`` to
+:func:`load`, :func:`loads` or :class:`CBORDecoder` to decode ordinary byte strings as
+:class:`bytearray` instead::
+
+    mutable_data = loads(b'\x44data', mutable_bytes=True)
+    mutable_data[0] = ord('D')
+
+Byte strings in immutable positions, such as map keys and semantic tag payloads that require
+bytes, remain :class:`bytes`.
+
 Date/time handling
 ------------------
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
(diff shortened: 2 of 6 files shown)


agronholm:
Why?

be-student:
The motivating use case is the one reported in [#266](https://github.com/agronholm/cbor2/issues/266): receiving large image byte strings and constructing a writable NumPy view with `np.frombuffer`. Returning immutable `bytes` requires the caller to create a mutable buffer first. The proposed flag makes mutable output opt-in while keeping map keys and semantic-tag contexts immutable.

That is the issue reporter's use case, not a firsthand application claim from this contribution. No performance benchmark is claimed; the older issue's exact copy-count analysis should not be assumed to describe the current Rust implementation.

be-student:
Fixed an interaction with string references: a byte string first decoded as a mutable value could later fail as a map key or bignum payload, while a key referenced as an ordinary value incorrectly remained immutable. The namespace now retains an immutable snapshot, and each reference follows its destination context with an independent mutable buffer; explicit shared-value references still preserve identity.

Added regression coverage for buffer/file decoding, hook-time mutation, and reference identity, and updated the documentation. Seven new cases failed before the fix; the complete suite now passes (480 tests), as do the pinned Ruff checks, Rust formatting, and type checking of the changed test file.

PR #340, opened 2026-09-09
fix integer overflow when decoding epoch-form dates (tag 100)

### LINKED ISSUE #305: Confusing decoding of standalone break marker; `allow_indefinite=False` still accepts it

sahvx655-wq:
## Changes

Fixes #305 in spirit but for a separate code path: while generating tag-100 test cases I hit a hard panic out of the decoder rather than a `CBORDecodeError`. `decode_epoch_date` reads the tag payload with `value.extract::<i32>()` and then adds `719163`, the ordinal of the Unix epoch. Any payload in roughly `[i32::MAX - 719163, i32::MAX]` overflows that `i32` addition, so under overflow-checked builds (debug, and the profiles the test/fuzz jobs run) the decoder aborts with `attempt to add with overflow` at `decoder.rs:1307`, surfacing to Python as a `pyo3_runtime.PanicException`. Under release the addition wraps instead, which happens to land out of `date.fromordinal`'s range and errors, but the arithmetic is still wrong by construction and the behaviour differs between build profiles.

Root cause is the premature narrowing: the payload is an unbounded number of days from the epoch, and the old pure-Python implementation added the offset on a Python `int` and let `date.fromordinal` reject anything out of range. The risk of leaving it is a decoder panic reachable from a single crafted tag on untrusted input, exactly the class of thing the fuzz harness is meant to catch. The fix keeps the addition on the Python object (`value.add(719163)`) so no Rust integer is involved, and `date.fromordinal` now raises a clean, wrapped decode error for out-of-range ordinals on every build profile. Valid dates and the full `date.min`..`date.max` round-trip are unaffected.

## Checklist

### DIFF
--- a/rust/decoder.rs
+++ b/rust/decoder.rs
@@ -1303,10 +1303,6 @@ impl CBORDecoder {
 
     fn decode_epoch_date(value: Bound<PyAny>, _immutable: bool) -> PyResult<DecoderResult> {
         // Semantic tag 100
-        // The payload is an unbounded number of days from the Unix epoch, so add the epoch's
-        // ordinal on the Python object rather than narrowing it to an i32 first: a payload near
-        // i32::MAX otherwise overflows the addition (a panic under overflow-checked builds).
-        // date.fromordinal() rejects genuinely out-of-range ordinals with a clean error.
         let py = value.py();
         let ordinal = value.add(719163)?;
         DATE_FROMORDINAL
--- a/rust/decoder.rs
+++ b/rust/decoder.rs
@@ -1303,9 +1303,16 @@ impl CBORDecoder {
 
     fn decode_epoch_date(value: Bound<PyAny>, _immutable: bool) -> PyResult<DecoderResult> {
         // Semantic tag 100
+        // The payload is an unbounded number of days from the Unix epoch, so add the epoch's
+        // ordinal on the Python object rather than narrowing it to an i32 first: a payload near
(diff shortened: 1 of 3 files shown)


rust/decoder.rs
```diff
@@ -1303,9 +1303,16 @@ impl CBORDecoder {
 
     fn decode_epoch_date(value: Bound<PyAny>, _immutable: bool) -> PyResult<DecoderResult> {
         // Semantic tag 100
+        // The payload is an unbounded number of days from the Unix epoch, so add the epoch's
+        // ordinal on the Python object rather than narrowing it to an i32 first: a payload near
+        // i32::MAX otherwise overflows the addition (a panic under overflow-checked builds).
+        // date.fromordinal() rejects genuinely out-of-range ordinals with a clean error.
```
agronholm: Please, no historical AI babble here.
```suggestion
```

sahvx655-wq: Agreed, that comment was more backstory than the code needed. The one-liner plus the changelog entry cover it, so the trim is right.

PR #341, opened 2026-09-09
reject misplaced break stop code instead of leaking break marker

### LINKED ISSUE #305: Confusing decoding of standalone break marker; `allow_indefinite=False` still accepts it

sahvx655-wq:
## Changes

Fixes #305.

**Break stop code leaking out of the decoder as its internal marker**

While generating interoperability test cases the reporter noticed that `loads(b"\xff")` returned an opaque sentinel object rather than raising, and traced the same object leaking wherever a `0xff` byte sat in a data-item position: as an element of a definite-length array, as a key or value of a definite-length map, as a tagged value, and at the top level. Walking the decode loop confirmed the cause. Major type 7 / additional information 31 is decoded into the internal `break_marker` and handed back as a normal value; only the indefinite-length array and map callbacks test for it and stop, so every other frame stored it verbatim. Per RFC 8949 section 3.2.1 a break appearing anywhere other than directly closing an indefinite-length string, array or map makes the enclosing item ill-formed, so none of these should have decoded at all. `allow_indefinite=False` did not help either, since that flag only guards indefinite *lengths*, not the standalone stop code.

The fix marks the two frames that may legitimately be terminated by a break (the indefinite-length array and map) and rejects the break marker with a `CBORDecodeError` anywhere else, both when it is delivered to a non-accepting frame and when it is left standing at the top level. Left unfixed this is a correctness and robustness problem for anything decoding untrusted CBOR: a sentinel with no public identity flows into decoded containers and tag values, so callers cannot tell malformed input apart from real data and may store or compare against an object they have no way to name. Existing indefinite containers, strings and the empty-container cases are unaffected; the previously leaking inputs now raise. A regression test covers the top level (with and without `allow_indefinite`), definite arrays and maps, nested arrays, tags, sets and the string-namespace tag.

## Checklist

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,11 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed the decoder returning its internal break marker as a value when a break stop code (major
+  type 7, additional information 31) appeared where a data item was expected, such as at the top
+  level, inside a definite-length array or map, or as a tagged value. Such input is ill-formed per
+  RFC 8949 section 3.2.1 and is now rejected with a :exc:`CBORDecodeError`
+  (`#305 <https://github.com/agronholm/cbor2/issues/305>`_; PR by @sahvx655-wq)
 - Fixed :class:`CBORSimpleValue` hashing inconsistently with the integer it compares equal to
   (`#334 <https://github.com/agronholm/cbor2/pull/334>`_; PR by @sahvx655-wq)
 - Fixed canonical encoding with value sharing enabled emitting shared references to container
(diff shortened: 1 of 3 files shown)


sahvx655-wq:
This went conflicting once #343 landed, since both patches touch the `BeginFrame` arm in `decode()`. I've merged master in and kept both sides: the `accepts_break` flag sits alongside the new `shareable_index` handling.

I rechecked the interaction rather than trusting the textual merge. A break in a shareable position (`d81cff`) is still rejected, an indefinite-length array inside a shareable still decodes and resolves through tag 29, and the full suite passes locally and in CI.

PR #343, opened 2026-09-20
make shared references honour object_hook and tag_hook results

sahvx655-wq:
## Changes

A shared reference (tag 29) to a map or an unhandled tag resolves to the raw `dict` or `CBORTag` rather than what `object_hook` or `tag_hook` returned. I hit it diffing decode output against 5.6.5 with hooks installed: `loads(bytes.fromhex("82d81ca1616101d81d00"), object_hook=Wrapped)` prints `[Wrapped({'a': 1}), {'a': 1}]` on master, while 5.6.5 returns the same `Wrapped` object twice. The root cause is that the `decode()` loop stores a mutable container in the shared value table when its frame begins, which cyclic references need, and never refreshes the slot on completion, so a sender can use tags 28/29 to hand the application values that skipped the hook it relies on for conversion or validation.

The finished value is written back to the slot when the frame completes, which is what #48 did for the 5.x decoder. The container frame has to inherit the marker's index for that, hence the extra `decoder_callback.is_none()` condition: without it a child container pops its shared parent. The new tests fail on master for `object_hook[mutable]` and both `tag_hook` cases; `7.0` carries the same loop, and I targeted master because this restores documented behaviour, but say if you would rather have it there.

## Checklist


agronholm:
Thanks!

